PR #127, opened 2026-06-07
feat(verify): server route verifies Rekor inclusion from carried proof (#119 P1)

npstorey:
**P1 consume (server side)** — pairs with typedstandards#8 (browser verify-flow). Follows the carrier-produce #126 (merged) and verify-core **0.2.0** (published).

## What
The server verify route now threads the carried Rekor inclusion proof + entry body into `verifyEvidence`, so §9.2 **check #8 deepens from hash-PARITY to cryptographic Merkle inclusion** against the pinned log key — verified **offline** from the carried body (no re-fetch).

- Parse `base_package_rekor_inclusion_proof`; only a real proof (audit path + signed checkpoint) is used. Empty `{}` early rows → treated as absent (healed by the backfill).
- Pass `rekorInclusionProof` + `rekorEntryBody` to `verifyEvidence`; bump `@typedstandards/verify-core` → `^0.2.0`.
- Surface `result.rekorInclusion` as a **new, additive** response field.

## Guardrails
- **Additive only** — existing response fields (`rekorVerified` hash-parity, etc.) unchanged; **#8 UI/depth labels NOT touched** (surfacing is P4). Data threading only.
- tsc/lint/build green (pre-existing repo errors only); full suite **285/285**.
- **Prod parity intended to deepen**: after deploy, 255b8e's `/verify` gains a `rekorInclusion` verdict; da9246 (no Rekor) stays null. Re-baseline — not a regression.

## After merge
Run the self-validating **backfill** (migration 0011 already applied) to heal existing rows' entry body + empty `{}` proofs, then re-confirm 255b8e.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
         "@neondatabase/serverless": "^1.0.2",
         "@noble/curves": "^2.2.0",
         "@noble/hashes": "^2.2.0",
-        "@typedstandards/verify-core": "^0.1.0",
+        "@typedstandards/verify-core": "^0.2.0",
         "@vercel/blob": "^2.3.3",
         "@vercel/kv": "^3.0.0",
         "@vercel/sandbox": "^1.10.2",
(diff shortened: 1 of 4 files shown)


npstorey:
Backfill script added to this PR: `scripts/backfill-rekor-entry-body.ts`.

Self-validating — it only writes a body that **folds to a verified inclusion proof** (against the already-stored proof when real; against the entry's fresh proof when the row stored an empty `{}`, refreshing both columns). Rows that fail to fold are skipped and reported, never written. Idempotent; dry-run by default.

Run after merge (migration 0011 already applied), then re-confirm 255b8e:

```
op run --env-file=<your local env> -- npx tsx scripts/backfill-rekor-entry-body.ts          # dry run
op run --env-file=<your local env> -- npx tsx scripts/backfill-rekor-entry-body.ts --apply  # write
```

Issue #258, opened 2026-08-13
Signed-output identity requires explicit instance configuration — remove the reference-identity defaults (portability charter 1)


npstorey:
## Origin

Charter 1 of the portability program formed from the civic-ai-tools#148 audit (full record + rulings: https://github.com/npstorey/civic-ai-tools/issues/148#issuecomment-5281912416). Findings A1, A2, A9/C4, C5, E1, with E2/E3/E13 as riders. **Wave 2** — boots after the owner-run env pre-step (below) is confirmed done.

## Scope

| id | location | task |
|---|---|---|
| A1 | `src/lib/site-config.ts:148–173` | The `DEMO_*` identity set (signer identity, publication host, trust-registry URL pair, platform title) stops defaulting into signed output. Shape: **required-or-refuse** on any path that signs; the unsigned tier (`resolveUnsignedIndicator`) is the honest-absence state. No "undeclared" sentinel (rulings record). |
| A1 | `docs/deploy.md:494–499` | Reclassify: the identity set moves out of "merely overrides a default" into the required-for-signing tier. `scripts/preflight-env.mjs:245–254` moves the set accordingly (grouped with the signing pair's relationship note at `:351`). |
| A2 | `src/lib/notebook.ts:97,163` | Attribution getters are client-bundled, so server config never reaches them. Restructure to thread the server-resolved values through (the `citation-placeholder.ts` / `EvidenceOriginProvider` pattern already in-tree). |
| A9 + C4 | `src/lib/mcp/registry.ts:106`, `compare-stream/route.ts:98` | `SOCRATA_MCP_URL` loses its reference-host fallback: unconfigured → fail loudly (G0: at boot vs at query). The signed-trace field (A9) then records only explicitly-configured routing. |
| C5 | `ChatNotebookOutput.tsx:88` | Resolve the `NEXT_PUBLIC_SOCRATA_MCP_URL` / `SOCRATA_MCP_URL` name split — one configured value must reach both server and client surfaces. Note: no `NEXT_PUBLIC_` var exists in `.env.example` today. |
| E1 | `public/robots.txt` + `layout.tsx:73` | Indexing becomes explicit instance config. Reference **keeps noindex for now** by setting the flag; the shipped default stops being a silent pre-launch leftover. Document the file. |
[...]

Issue #30, opened 2026-03-10
Direct API calls to Anthropic/OpenAI/Google instead of OpenRouter


npstorey:
## Idea

Replace OpenRouter with direct API calls to each provider to reduce cost overhead and latency.

## Current state

All model calls go through OpenRouter (`https://openrouter.ai/api/v1`), which adds a markup (typically ~5-15% on token costs) and an extra network hop. This was the right choice early on — single integration point, easy model switching, unified billing. But as usage grows, the overhead adds up.

## Proposal

Add support for direct provider API calls:
- **Anthropic**: `@anthropic-ai/sdk` → `api.anthropic.com`
- **OpenAI**: `openai` SDK → `api.openai.com`
- **Google**: `@google/generative-ai` SDK → `generativelanguage.googleapis.com`

### Architecture options

1. **Provider abstraction layer**: A unified interface that routes to the correct SDK based on model ID. The rest of the codebase stays the same.
2. **Configurable backend**: Environment variable (`API_BACKEND=direct|openrouter`) so the site can switch between modes. Keep OpenRouter as a fallback for models not available via direct API.
3. **Hybrid**: Use direct APIs for the default/recommended models, OpenRouter for everything else.

### Considerations

- Requires 3 API keys instead of 1 (more env vars, more key management)
- Each provider has slightly different tool/function calling formats — need an adapter
- OpenRouter handles rate limiting and fallbacks; direct calls need their own retry logic
- Billing is split across 3 dashboards instead of 1
- For the eval script and local CLI, users would need to provide per-provider keys

### When to do this

Not urgent while usage is low. Worth doing when:
- Monthly OpenRouter spend exceeds ~$50/month (overhead becomes meaningful)
- Latency becomes a user complaint (extra hop matters for streaming)
- A provider offers features not available through OpenRouter

## Related

- website#24 (model evaluation)
- src/lib/openrouter.ts (current API client)

npstorey:
**Ruled 2026-08-19 (NYC decision round, recorded by the CAT PLAN coordinating seat): needed — this graduates from idea to scheduled scope (Wave N4 of the pre-deployment program).**

Two updates to the issue's framing:

[...]

PR #311, opened 2026-08-24
feat: the wire string and the recorded identity separate (#30 P3)

npstorey:
Phase P3 of the gated sprint on #30 (Wave N4 — endpoint-generic model layer). Base `13983d1` (`rollback/pre-30-p3`), branch `sprint-30/p3-identity`. **IMPL model: Opus 5.** Worked in an isolated git worktree; the shared checkout was never touched.

## What this does

P2 built the catalog that carries both strings and deliberately did **not** thread `endpointModel` onto the wire, because doing so before the declared identity had a home would have pushed an operator's deployment alias into signed output. This performs the split.

Call sites now carry a `ModelIdentity` — `{ endpointModel, declared }` — instead of one string doing two jobs. The wire gets `endpointModel`. `cost.model`, `environment.modelVersion`, the PROV-O model agent's `dcterms:title`, `analysis.model`, `gen_ai.request.model`, the notebook's author stamp and an evaluation attestation's `methodology.evaluatorModel` all get `declared`.

The strongest structural guarantee here is one line of interface design: **`packager.ts` is given no name for the wire string at all.** `PackageInput.model` is documented as the declared identity and there is no second field, so "no deployment alias in the package" is a property of the packager's interface rather than of every call site's discipline.

## The four announced byte changes

Under the default endpoint the two strings are identical, so nothing about the reference instance's *behaviour* changes. Four things about its *bytes* do, all of them announced:

**1 — E4, the PROV-O model agent's description.** The only packager-level change, and the one criterion 1 names.

| | |
|---|---|
| before | `"dcterms:description": "Large language model via OpenRouter"` |
| after (default dialect) | `"dcterms:description": "Large language model reached over an OpenAI-compatible chat-completions API"` |
| after (azure dialect) | `"dcterms:description": "Large language model reached over a deployment-routed chat-completions API"` |

The old value was spread out of the harness's `CIVICAITOOLS_PROVENANCE_CONFIG` — the reference deployment's own gateway, asserted by every instance including instances that have never touched OpenRouter. It is the same spread-a-reference-constant shape #258 and #294 removed from the platform agent, on the one remaining field that is an endpoint fact rather than a vocabulary fact. `sourceRegistry` stays spread; that one is shared vocabulary.

The phrasing names no vendor product, no resource host and no deployment. What is left is the only part a reader can act on: one dialect addresses a model by slug at a shared path, the other addresses a deployment by name.

**2 — E6, the notebook's reproducibility stamp.** Section E of a datHere package, so inside what the content hash fingerprints.

| | |
|---|---|
| before | `- **Author model:** anthropic/claude-sonnet-4-6 via OpenRouter.` |
| after | `- **Author model:** anthropic/claude-sonnet-4-6, reached over an OpenAI-compatible chat-completions API.` |

**3 — E5, `gen_ai.response.model`.** New traces gain one attribute per `llm_inference` and `synthesis` span, carrying what the endpoint itself reported. Historical records are frozen and are not re-emitted.

**4 — absent usage is absent.** `cost.promptTokens` / `cost.completionTokens` are omitted rather than recorded as `0`, extending the `totalTokens || undefined` idiom that was already there. Real counts are untouched.

Deltas 2–4 change the *inputs* the packager is given for newly captured records; delta 1 is the packager's own output. `packager.test.ts:257` and `:321` are green, untouched.

## Acceptance

| # | Criterion | Where |
|---|---|---|
| 1 | Reference bytes do not move except E4 | `packager.test.ts` — `:257`/`:321` untouched and green; the E4 delta pinned with its exact string, plus an assertion that `OpenRouter` no longer appears |
| 2 | Declared identity reaches all three sites; wire gets `endpointModel` | `model-identity.test.ts` — under a fake-Azure fixture where the two differ, every captured request carries `example-deployment-alias` and never `vendor/model-1`; `cost.model`, `environment.modelVersion` and the PROV title all carry `vendor/model-1` |
| 3 | Both strings in one signed artifact; mismatch warns, does not gate | `model-identity.test.ts` — the span carries `gen_ai.request.model` (declared) beside `gen_ai.response.model` (reported); the mismatch warns server-side with both values and the query still completes with no stream error. A second case proves an agreeing endpoint records without warning |
| 4 | No provider or infrastructure identifier in the package | `model-identity.test.ts` — asserted over the **serialized canonical JSON**, not field by field: the deployment alias, the resource host, the port, the api-version, the credential and `azure`/`Azure` are all absent |
| 5 | Absent usage absent, not zero; `include_usage` under the OpenAI dialect | `packager.test.ts` (canonical `cost` reduces to `["model"]`), `model-identity.test.ts` (the streamed request body carries `stream_options: {include_usage: true}`, and the azure dialect deliberately does not) |
| 6 | The routed bypass closed, CI green | `generate-summary` resolves through the catalog; full check set below |

## The routed defect (§2.7), closed by role rather than by literal

`SUMMARY_MODEL = 'google/gemini-3.5-flash-lite'` is gone. The catalog gains a `summarizer` role beside `default` and `evaluator`; the built-in entry for that same id carries it, so this instance calls exactly what it always did. What changes is that the choice is now instance-configuration, visible to catalog validation, and cannot name an id the instance does not offer. A catalog declaring no summariser falls back to its default — the draft is a convenience the publisher edits before anything is signed, so it is not on the evidence path and refusing to publish over it would be out of all proportion. Two claimants are refused, both named.

## `gen_ai.system`, measured rather than assumed

`CIVICAITOOLS_TRACE_CONFIG` declares `semconvVersion: '1.30.0'`. Read against the 1.30.0 attribute set: `az.ai.openai` and `openai` are registered values of `gen_ai.system`; **`openrouter` is not, and never has been** — it was already the emitted value before this phase.

So the mapping takes a registered value where one exists and keeps the unregistered one only where semconv offers nothing: azure → `az.ai.openai`; any non-default OpenAI-compatible base URL → `openai`; the built-in OpenRouter endpoint → `openrouter`, unchanged. Semconv's own rule for a system with no registered value is the provider name in lowercase, which is what that is — and calling a router `openai` would be less honest, not more, as well as an unannounced byte change to the reference instance.

**No `server.address` was added**, deliberately.

## Two correctness consequences of the split, fixed here

- **Evaluator independence was comparing two namespaces.** The analysis model is only ever read from `pkg.cost.model`, a *declared* identity, and it was being compared against catalog *ids*. Under a catalog where they differ the gate would have picked the analysis model to grade its own work. Both the publish route's override check and `resolveEvaluatorModel` now compare declared-against-declared.
- **Replay would have sent a string no endpoint answers to.** `pkg.cost.model` is a declared identity; the replay path maps it back through `endpointModelForDeclared`, carrying it through unchanged when no entry declares it — which is what a record naming a model this instance no longer offers has always needed.

## The P2 invariant test, updated deliberately

`model-catalog.test.ts`'s `CATALOG: an omitted model under the default dialect declares the endpoint string` is kept, with all three assertions intact and its reasoning rewritten. P2 wrote the third assertion as a **licence**: its call sites kept sending `entry.id` on the wire, and this test was the standing proof that doing so was harmless because under the built-in catalog the two are the same string. Its comment said that if it ever failed, the split was overdue rather than the test wrong.

The split has happened, so the licence is spent and the test's job changes rather than ending. The same assertions now pin that the built-in catalog is the case where the split is a **no-op**, and therefore that the reference instance's canonical JSON does not move. A fourth assertion was added on the pair itself. Nothing was deleted.

## Blast zone

26 files, +1236/−78, all under `src/`. Beyond the contract's named list, five extensions, each because the phase is otherwise incoherent — named here rather than left to be discovered:

- `src/lib/model-client.ts` — `getGenAiSystem()` and `includeStreamUsage()`. Both are dialect questions and the dialect lives here; deriving them anywhere else would be a second copy of the closed set.
- `src/lib/openrouter.ts`, `src/lib/notebook-author/synthesize.ts`, `src/lib/evidence/adversarial-eval.ts` — call sites that pass a model to the client or to the notebook, which the contract's blast zone names as a class.
- `src/app/api/evidence/route.ts` — the site where identity actually enters the package. Without it, criterion 2 holds only for direct packager calls and not for a single real publish.
- Six further routes that pass a model to the client (`compare`, `compare-stream`, `query-notebook`, `publish`, `evaluate`, `replay`), plus five test/dev fixtures that construct a notebook.

## Checks

`npm ci` first. All run on macOS, Node 22.23.1, in the worktree.

```
$ npm test
1..922
# tests 955
# suites 10
# pass 955
# fail 0
# cancelled 0
# skipped 0
# todo 0
# duration_ms 1521.851958

$ npm run build
▲ Next.js 16.3.0 (Turbopack)
✓ Running next.config.ts took 10ms
✓ Compiled successfully in 246ms
  Running TypeScript ...
  Finished TypeScript in 958ms ...
✓ Generating static pages using 9 workers (32/32) in 178ms
(2 pre-existing Turbopack warnings, then the full Route (app) table)

$ npm run typecheck
(no output, exit 0)

$ npm run lint
(the documented baseline: eval-models.mjs, AttestationDialog.tsx, RenderingCellOutputs.tsx, bpmn/animation.ts)

$ npm run check:compose-env
```

**Build note:** `npm run build` succeeded on Turbopack **in this sandboxed agent session** — it did not need `--webpack`. That contradicts nothing: `CLAUDE.md` says to re-measure rather than assume, and the measurement this session came out the other way from P1's and P2's. CI's unsandboxed job is the real gate.

## No new environment variables

`summarizer` is a catalog field, not a variable. No `ENV_SPEC`, compose or `.env.example` rows are needed, so this gate has no owner append.

---

## Contract premises — what survived and what did not

Every quoted command output in §2 was re-run at `13983d1` and reproduced exactly, including all eight line numbers. This contract was measurably better-measured than P1's and P2's. Four findings, none of them a false line number:

**1 — §2.1's "three sites where identity enters the package" is four.** The fourth is `methodology.evaluatorModel` inside a signed `attestation/evaluates/v1` node (`adversarial-eval.ts`). P2's own design note named it — "the wire parameter AND the identity that lands in `analysis.model`, in the notebook stamp, and in an evaluation attestation's methodology" — but §2.1 counted only the three inside `packager.ts`. Handled here: `runAdversarialEval` takes the pair and the attestation gets `declared`. Left unhandled it would have been the one signed surface still asserting a deployment alias.

**2 — §2.3's semconv question, answered by measurement.** The declared version is `semconvVersion: '1.30.0'` (`capture/trace.ts:118` in the harness). Against the 1.30.0 attribute set, `az.ai.openai` and `openai` are registered `gen_ai.system` values and `gen_ai.response.model` is a registered attribute — so nothing needed minting. But **`openrouter` is not registered**, and that is the value already in the tree at HEAD. Reported rather than silently changed, per the rider; kept, per semconv's own fallback rule and per criterion 1.

**3 — §2.5's omission is only visible in canonical form.** The `totalTokens || undefined` idiom is exactly as quoted, but produce-core leaves the key present holding `undefined`; `JSON.stringify` and `canonicalize` are what drop it. `Object.prototype.hasOwnProperty` on the in-memory package still answers `true`. The test therefore asserts over the canonical bytes — which is what the hash, the signature and a reader's download all see. An assertion written against the object would have failed while the behaviour was correct.

**4 — §2.6's quoted grep output is one line short.** `grep -n "^export function" src/lib/model-resolver.ts` returns six matches at `13983d1`, not five: `_resetModelCatalogForTests` at `:178` is missing from the quote. Not load-bearing; recorded for completeness.

**On criterion 1's wording.** "Canonical JSON is byte-identical except E4's description" is exactly true at the packager level — same inputs, same bytes. E5 and E6 change the *inputs* for newly captured records, and both are required by §2.3 and §2.4 of this same contract, so the criterion and the tasks are consistent once read that way. Stated explicitly above as four announced deltas rather than left to inference.

**On E4's two permitted shapes.** §2.2 allowed either a dialect label or omission. The label was chosen: omission would have deleted reader-facing information to solve a problem that a derived value solves without loss, and it would have made E4 a deletion rather than a change that does something under Azure.

## Flagged, not fixed

- **The trace still records `0` for absent token usage.** `gen_ai.response.prompt_tokens` / `gen_ai.response.completion_tokens` use `|| 0` in all four `endSpan` calls in `openrouter-streaming.ts`. This is the same false-precision defect criterion 5 fixed in `cost.*`, one layer down and also inside the signature. Not fixed here: criterion 5 named `cost.*`, and changing it moves trace bytes for every new record with no criterion asking. Worth a named criterion in a later phase.
- **`POST /api/evidence` does not validate that `body.model` is a non-empty string.** Pre-existing; the change here preserves the prior behaviour exactly (an absent value records as absent). A 400 belongs with the rest of that route's validation.
- **`/api/compare` and `/api/compare-stream` still forward an unoffered model id upstream.** P2's refuse-before-upstream criterion covered only the notebook route, and these two are resolved tolerantly here so that no new refusal ships. Making them strict needs `QueryForm.tsx:99`'s hardcoded initial state fixed in the same pass — already a P4 flag.
- **`AttestationDialog.tsx`'s stale `EVALUATOR_MODELS`** (P4, as routed) is now slightly more consequential: the evaluation-preview route resolves that id tolerantly, so a dialog id absent from the catalog reaches the endpoint unresolved under a non-default dialect.
- **Two statements this change falsifies, outside the zone:** `scripts/eval-models.mjs:6` and `docs/project-plan.md:265` still say "via OpenRouter". P4's statements pass. (`src/app/(marketing)/talks/_decks/…:1359` is a frozen talk deck describing what the code did in July — correct as history, deliberately untouched.) The one in-file comment this change falsified, `adversarial-eval.ts`'s "Run the rubric against a package via OpenRouter", is fixed here rather than left standing in a file the diff already opens.

## One behaviour change the gate should accept explicitly

`buildEvidencePackage` now reads `getModelApiKind()`, which **refuses an unrecognized `MODEL_API_KIND`**. An unset value still takes the default, so only a typo is affected — but on an instance whose only publishing path is the Claude Code skill (which never calls the model endpoint), a typo'd kind that used to be harmless now blocks publishing. That is ADR-0024 §A applied consistently: on the evidence path a value is present or it is an error, and a package must not guess how its model was reached. Recorded here rather than discovered later.

## One ask on P4

`summarizer` is a new catalog field. P4's `docs/instance-setup.md` live-wiring checklist and its catalog example should carry it, alongside `default` and `evaluator`.

## Gate state

Merge-ready at head `8ab81b7`. Two commits, both `git commit -s`, author and trailer `Nathan Storey <npstorey@users.noreply.github.com>` on each. `MERGEABLE` / `CLEAN`. Required checks at this head: `DCO` pass · `Vercel` pass · `build / test / lint / typecheck` pass (1m8s). No `.env.example` rows are needed, so this gate has no owner append. IMPL does not merge.


npstorey:
## Seat GO — P3 at `8ab81b7`

**Head-bound GO for PR #311 at `8ab81b77a440b528e353768a8e4cf47a354fe8bc`** (base `13983d1`, `rollback/pre-30-p3`). Owner: approve and merge this head. If the head moves, this GO lapses.

### Verified independently of the gate record (5394543652)

- **Numstat = the PR API.** Three-dot `origin/main...origin/sprint-30/p3-identity`: 26 files, +1242/−78, nothing outside `src/`. Two commits, both signed off.
- **Check rollup on the exact head:** `build / test / lint / typecheck` ✅ · `DCO` ✅ · `Vercel` (commit status, deployment completed) ✅ · `Vercel Preview Comments` ✅ — the ruleset's three required contexts all green. **`standalone build + asset check` did not run and that is by design, verified in the workflow itself:** `standalone.yml` triggers only on `Dockerfile`, `next.config.*`, `scripts/**`, `src/lib/notebook-author/helpers/**` and its own file; this diff touches none of them (its two notebook-author files sit outside `helpers/`).
- **Read in the code, not the report:**
  - **The split, at every wire site:** all five streaming calls (`openrouter-streaming.ts`), both `openrouter.ts` helpers, evaluate, replay and generate-summary send `endpointModel`; the packager's `model` input is documented at the interface as declared-only, so "no alias in the package" is a property of the module, not of call-site discipline.
  - **The fourth identity site:** `adversarial-eval.ts` takes `ModelIdentity`, sends `endpointModel` on the wire (`:77`) and hands `declared` to `emitEvaluationAttestation` for `methodology.evaluatorModel` — closed as recorded.
  - **Criterion 4 over canonical bytes:** `model-identity.test.ts` serializes the package and asserts the deployment name, resource host, port, api-version, credential, and the provider name (both cases) are absent — not a field-by-field check.
  - **Hash guards intact by construction:** `packager.test.ts` is purely additive (+70/−0), so `:257`/`:321` cannot have moved; no `*hash*`/`*canonical*`/`*sign*` test file is touched.
  - **§2.5 both halves:** the cost layer records absent usage as absent (`|| undefined`), and `includeStreamUsage()` asks for usage only under `openai-compatible` — the azure api-version gate is its own test. The span-level `|| 0`s remain, exactly as **#312** describes — **filed rather than routed is right** (P4 does not own the trace, and it moves signed trace bytes).
  - **The routed bypass closed properly:** `SUMMARY_MODEL` is gone; `generate-summary` resolves the new `summarizer` role (single-claimant validated, falls back to the default entry), 503 on a config error, and correctly records no identity — nothing it produces is signed.
  - **Reference-instance behaviour preserved where it must be:** the evaluate route resolves **tolerantly** (`modelIdentityForValue`), so the dialog's stale five-entry list — P2's flagged P4 item — keeps working; the publish override stays strict as P2 built it, now compared declared-to-declared; `POST /api/evidence` carries an uncataloged `body.model` through, so external publishers are unaffected.
- **Both riders held:** all eight quoted line numbers reproduced (the ORCH's own accounting, consistent with zero fix-on-top); worktree isolation confirmed from the shared checkout itself — on `main`, clean, throughout P3.

### Accepted with the gate

- **`buildEvidencePackage` refusing an unrecognized `MODEL_API_KIND`** — concur, including for an instance that publishes only via the harness skill: a typo'd dialect means the instance cannot say how its model was reached, and ADR-0024 §A forbids guessing. Declared here, which is where it belongs.
- **`summarizer` as a catalog role** — P4 documents it in the setup section's catalog example.

### One bookkeeping flag, not blocking

The defect series doesn't add up as stated: "the fifth defect — P1: two, P2: four, P3: one" is seven by its own arithmetic. The wave-close ledger should restate the series explicitly (per phase: what was falsified, what kind) — it feeds the post-N4 re-ask on ORCH model policy, so the count must be exact there.

### For P4's contract

Carries forward, all previously routed and restated in this gate record: the wire-field rename for caller keys · the dialog's static `EVALUATOR_MODELS` (now marginally more consequential) · `summarizer` + `MODEL_CATALOG` operator docs and the instance-setup checklist · the statements pass (`eval-models.mjs:6`, `docs/project-plan.md:265`, marketing copy) · `/api/compare*` strictness with `QueryForm.tsx:99` in the same pass, or an explicit decision to leave them tolerant.

Ledger accepted as recorded: Opus 5 · one pass · 6/6 (criterion 1 restated as self-contradictory, deltas announced with before/after) · 0 fix-on-top · 4 premises falsified.

— CAT PLAN (seat), 2026-08-24

Issue #315, opened 2026-08-24
Should the model catalog carry pricing at all? Operator-maintained prices go stale under a signature


npstorey:
Should the model catalog carry `pricing` at all? Operator-maintained prices go stale, and a stale price under a signature is a durable false figure.

## Where this came from

Wave N4 P2 gave the instance catalog a `pricing` field. Wave N4's cold read then found that `docs/instance-setup.md` claimed it "is what a record's cost estimate is computed from" — **which is false**: `builtInPricing()` (`src/lib/model-catalog.ts:363-367`) reads `BUILT_IN_CATALOG` and then `HISTORICAL_MODELS`, and never a configured catalog. The lookup key is the *declared* identity while a catalog is keyed by `id`, so the two namespaces do not even match.

Wave N4 P6 corrects the documentation. **This issue is the question the correction leaves open**, and it is not "wire it up."

## The question, as the owner put it

> Do we really want the pricing feeds in there? The pricing changes so frequently it might be hard to know exactly how to handle pricing fields.

## Why "just wire it up" is the wrong default

A cost estimate computed from operator-supplied pricing lands in a **signed record**. Model prices change often and nothing in this system keeps an operator's catalog current — so the natural failure is a package asserting a dollar figure derived from a price that was wrong when it was signed. That is the same class as #312 (a trace asserting zero tokens when usage was merely absent): **a number under a signature that nothing keeps true.**

The built-in table has the same staleness problem, but it has a maintainer, a review path, and a single place to fix. A per-instance field has none of those and multiplies the surface by the number of instances.

## Options worth weighing

1. **Drop `pricing` from the catalog schema.** Cost estimates stay built-in-only; a configured instance gets no estimate rather than a possibly-wrong one, which is the omission-over-assertion posture this project already applies elsewhere. Costs: an operator on their own endpoint never sees a cost line.
2. **Keep it, scoped explicitly to display and kept out of anything signed.** The field feeds a UI estimate, never `cost` in a package. Costs: two pricing paths to keep straight, and the distinction has to survive future edits.
3. **Keep it and wire it into signed estimates, with staleness made visible** — a required `pricingAsOf` date, and the estimate omitted or marked when the date is old. Costs: the most machinery, and it puts a date an operator maintains into the same position.
4. **Status quo** — the field is accepted, validated, and never read. Now honestly documented as such by P6. Costs: a field in configuration that does nothing is a standing invitation to assume it does something, which is exactly the defect P6 is fixing.

[...]

Issue #314, opened 2026-08-24
Explore's live query sends a model id the catalog never resolves (McpFlowDiagram.tsx:15)


npstorey:
The explore page's live query sends a model id this instance does not offer — on every instance, including the reference deployment.

## What it is

`src/components/explore/McpFlowDiagram.tsx:15`:

```js
const DEFAULT_MODEL = 'anthropic/claude-sonnet-4';
```

That is **not a diagram label.** It is passed to `liveTrace.start(query, DEFAULT_MODEL, DEFAULT_PORTAL)` at `:145` and `:148`, and reaches `POST /api/compare-stream` as the `model` field. It is a wire value.

And that id is not in the catalog. Wave N4 P2 placed `anthropic/claude-sonnet-4` in `HISTORICAL_MODELS` (`src/lib/model-catalog.ts:219`) — the constant that exists to render **already-published records** and is deliberately **never resolvable**. So `/explore`'s live query names a model the instance does not offer.

## Why it has not been visible

`/api/compare` and `/api/compare-stream` resolve a caller's id **tolerantly** — a deliberate, recorded decision across N4 P3 and P4 — so the id is forwarded upstream rather than refused. Under the built-in OpenRouter endpoint it is a real slug and the query succeeds, which is why nothing has failed.

It stops being invisible under any non-default endpoint: an id absent from the catalog cannot be mapped to a wire model, and `/explore`'s headline feature — the live trace — is the thing that breaks.

## Why N4 P4 did not fix it

The correct replacement is a **product question**, not a lookup. The closest true analogue is the catalog's `default` entry, but that entry is `selectable: false` and therefore absent from `GET /api/models`. So a correct fix needs one of:

- a decision that `/explore` uses the **first offered** model (what `QueryForm` now does after P4 removed its own hardcoded initial state), or
- a new API surface exposing the default entry to clients, or
- an explicit `explore` catalog role, alongside `default`, `evaluator` and `summarizer`.

P4 was the last phase before a cold read and no criterion covered this, so it was flagged rather than widened into. It is the largest thing that phase found and did not close.

## History worth carrying

This was flagged during P2 as **"a diagram label"** — display-only — and that characterization propagated into P4's contract as settled. It was wrong. Worth remembering that a flag's *description* ages the same way a line number does, and re-reading the code rather than the flag is what caught it.

## Related

- civic-ai-tools-website#30 (Wave N4) — the catalog, the resolver, and the tolerant-resolution decision.
[...]

Issue #283, opened 2026-08-19
QueryForm: a 200 response without a models key throws at render


npstorey:
## What this is

`src/components/QueryForm.tsx` sets `setModels(data.models)` inside its fetch `try`, but the `catch` covers only fetch/JSON failures. If `/api/models` returns 200 **without** a `models` key, `models` becomes `undefined` and `models.find(...)` throws at render. Latent today (the route currently always returns the key), but the failure mode is a render crash rather than anything a reader can act on.

## Disposition

Make the failure loud-and-friendly rather than a throw: treat a 200-without-models as the same class as a failed fetch (the form-level error path), never a render-time exception. Same finding class as the compose acceptance's silent-hang: a well-formed-but-wrong response deserves an explicit failure mode.

**Rider on #30:** the endpoint-generic provider layer rebuilds exactly this path — whoever implements #30 should land this guard as part of that rebuild rather than patching the current code twice.

## Provenance

Found during Wave N1 P2's premise-check (#275, PR #279 — recorded in the P2 gate record on #179 as an out-of-scope observation, deliberately left untouched by that phase's one-focused-change rule). Filed at the Wave N1 close-out per the coordinating seat's disposition, 2026-08-19.

🤖 Filed by the Wave N1 ORCH session at close-out.

npstorey:
# Wave N5 P2 gate record (#283, anchor #282) — loud-and-friendly model-list failure shipped

**PR #286** (`fix: route a shape-invalid /api/models response through the form's error state (#283)`), head `557d7eda5c3db98a1df313213b7434f0e4f88e75`, squash-merged to main as `87ac79928b69d16a6a345657a9d43deab5232091` under the two-key gate: coordinating-seat GO (head-bound, independently patch-validated; merge window granted first-priority) + the owner's second key exercised by merging the PR directly on GitHub (`merged_by npstorey`, 2026-08-20T11:16:24Z). Head binding verified post-hoc per the standing control: headRefOid at merge == the GO'd head, ORCH-verified from the API, seat flagged for its own verification. Tags: `rollback/wave-n5-p2` (pre-phase, at `91d4950`) and `rollback/wave-n5-p2-merged` (at the merge commit).

## Check rollup on the merged head (all pass, platform checks included)

- `build / test / lint / typecheck`: run [32362603334](https://github.com/npstorey/civic-ai-tools-website/actions/runs/32362603334) (1m18s)
[...]

PR #317, opened 2026-08-25
fix: /explore's model-fetch cache never clears on failure (#30 P7)

npstorey:
## Summary

P7 of #30 (Wave N4, last phase) — closes two defects the scoped re-read found in P6's own new code in `src/components/explore/McpFlowDiagram.tsx`.

**(A) A poisoned model-fetch cache.** `offeredModelRef` cached the fetch's own promise, including a rejected one; `.catch()` resolved it to `''` without clearing the ref, so one failed or empty `/api/models` response at mount poisoned every later `/explore` live-query click until a full page reload. Replaced with `createOfferedModelResolver` (`src/lib/offered-model.ts`), a plain factory that caches only a resolved, non-empty id — a failure leaves nothing cached, so the next `resolve()` retries.

**(B) Two inaccurate comments from the previous phase**, corrected to what I verified rather than what was claimed:
- `/explore`'s live trace is **not publishable**: `McpResponseDisplay.tsx`'s `canPublish` requires `!!evidenceTrace`; `LiveResponsePanel` renders `McpResponseDisplay` without an `evidenceTrace` prop; `useLiveTrace.ts` has no case for the `trace` SSE event `compare-stream/route.ts` emits (only `progress`/`token`/`complete`/`error`). The trace is recorded server-side and discarded on this surface.
- The route's "Query and model are required" 400 does **not** surface as something the reader can read — that body carries no `code`, so `classifyStreamError` falls through to `generic` copy. This is now moot on this surface: the client bails before ever sending that request.

The accurate versions of both claims at `compare-stream/route.ts:69-71` and `model-resolver.ts:256-259` are untouched (`git diff` confirms zero changes to either file).

## What changed

- **`src/lib/offered-model.ts`** (new) — the retry/caching policy, extracted into a plain factory so it's unit-testable by `node:test` (McpFlowDiagram.tsx is JSX, which `--experimental-strip-types` can't parse — the same reason `parseModelsResponse` was extracted into `model-list.ts` in #283).
- **`src/lib/offered-model.test.ts`** (new) — 5 tests, including the required "a failed fetch is not cached" case.
- **`src/components/explore/McpFlowDiagram.tsx`** — uses the resolver; tracks a `modelsError` flag (mirrors QueryForm's identically-named flag, P4) that clears on success and stays set on durable failure; `handleLiveStart` now bails before calling `liveTrace.start` when the resolved id is empty; corrected the two comments.
- **`src/components/explore/TraceControls.tsx`** — new `modelsUnavailable` prop; withdraws the Run control and discloses `MODELS_LOAD_ERROR` while it's true, rather than inviting a click that can't work (design-principles.md Principle 3's corollary: a list that failed to load is not a bad selection — nothing here is marked invalid).
- **`src/lib/streaming.ts`** — `MODELS_LOAD_ERROR` moved here (exported) from QueryForm.tsx (was private), per this repo's own rule for cross-cutting reader copy now that it has a second caller.
- **`src/components/QueryForm.tsx`** — one-line change: imports `MODELS_LOAD_ERROR` from `streaming.ts` instead of declaring it locally. No behavior change.
- **`src/app/api/model-refusal-ordering.test.ts`** — updated to check `/explore`'s fetch+parser across `McpFlowDiagram.tsx` and its new `offered-model.ts` dependency, since the extraction moved the literal `'/api/models'`/`parseModelsResponse` calls out of the component file this pre-existing test reads as source. This was a mandatory adjacent fix — the test failed against my refactor before this update (the behavioral property it protects, "no client hardcodes a model id," still holds; the implementation location moved).

### Blast-zone note

The contract named `McpFlowDiagram.tsx`, its test, `LiveResponsePanel.tsx` (if needed — it wasn't), and `streaming.ts` (if reader copy belonged there — it does) as the expected zone, explicitly inviting a flagged expansion if the zone proved too small (as three earlier phases in this wave did). It was: withdrawing the Run control required touching `TraceControls.tsx` (that's where the control lives), deduplicating `MODELS_LOAD_ERROR` required a one-line touch to `QueryForm.tsx` (per the repo's own cross-cutting-copy rule), and the new `offered-model.ts`/`offered-model.test.ts` pair required a corresponding update to the pre-existing `model-refusal-ordering.test.ts`. All four are flagged here.

## Verified code facts (acceptance criterion 3)

- `canPublish` (`McpResponseDisplay.tsx:426`): `isComplete && !!content && !!evidenceTrace && toolsCalled.length > 0`.
- `LiveResponsePanel.tsx:162-173` calls `<McpResponseDisplay ... />` with no `evidenceTrace` prop.
- `useLiveTrace.ts` switches on `type === 'progress' | 'token' | 'complete' | 'error'` only (lines 330/383/389/410) — no `trace` case.
- `compare-stream/route.ts:251` emits `{ type: 'trace', panel: 'withMcp', data: traceJson }`; `useStreamingComparison.ts:501` (used by `QueryForm`/`/ask`, not `/explore`) is the only consumer with a `case 'trace'`.
- `compare-stream/route.ts:36-40`: the "Query and model are required" 400 body is `{ error: 'Query and model are required' }` — no `code` field.
- `compare-stream/route.ts:69-71` and `model-resolver.ts:256-259` (the accurate route-level claims) — confirmed unchanged in this diff.

## What the reader sees on a durable model-catalog-load failure

The "Try your own" live-query form stays visible (query text isn't discarded), but the Run button is disabled and a `role="status"` message appears in `--caution` color: *"Couldn't load the list of AI models this site offers, so a query can't be sent right now. Refresh the page to try again."* — the same copy QueryForm already shows for the identical failure class. Nothing is marked invalid (no red border on the query input); this is a background-load disclosure, not a validation error.

## Checks (model: Sonnet 5, isolated worktree)

Base: `eadd284` (`rollback/pre-30-p7`). `npm ci` run first.

**`npm run build`** — exit 0, `✓ Compiled successfully`.

**`npm run typecheck`** — exit 0, no output.

**`npm test`** — exit 0:
```
# tests 1009
# suites 10
# pass 1009
# fail 0
# cancelled 0
# skipped 0
# todo 0
```
(1004 pre-existing + 5 new in `offered-model.test.ts`; `model-refusal-ordering.test.ts`'s existing test was updated, not added, so the count above isn't 1004+6.)

**`npm run lint`** — exit 0:
```
```
Matches CLAUDE.md's documented baseline exactly; all 4 warnings are pre-existing and unrelated to this diff (`scripts/eval-models.mjs`, `AttestationDialog.tsx`, `RenderingCellOutputs.tsx`, `bpmn/animation.ts`).

**`npm run check:compose-env`** — exit 0:
```
```

## Flagged, not fixed

- Initial implementation of the lazy-init used `useRef` with a null-check-then-assign pattern (the idiom the `react-hooks/refs` lint rule's own error message recommends); ESLint's `react-hooks/refs` flagged it as "Cannot access refs during render" anyway. Switched to `useState(() => ...)` lazy initializer instead — functionally identical (constructed once, setter never called again), passes lint, no behavior change. Noting this in case the rule's stance on that documented idiom is worth a separate look; out of scope here.
- #314's product question (should `/explore` have its own catalog role, or reach the `default` entry) — unanswered, as directed.
- #312, #315, the three preflight/app disagreements, the harness source registry, `POST /api/records` tolerance, CLAUDE.md's stale test count, anything moving signed-package bytes — untouched, as directed.

## Premises checked

Every quoted line/behavior in the contract was verified against this worktree's HEAD (`eadd284`) before editing; all matched. No premise failed to survive the check.

### DIFF
--- a/src/app/api/model-refusal-ordering.test.ts
+++ b/src/app/api/model-refusal-ordering.test.ts
@@ -203,20 +203,33 @@ test('#30 P6 F3: the guard reports the variable actually missing, not the key', 
 // /explore's headline feature into a 400 on every instance.
 
 test('#30 P6 F1 / #314: no query client names a model id in its own source', () => {
+  // website#30 P7 extracted /explore's fetch+parse into a plain module
+  // (`offered-model.ts`) so its retry/caching policy is unit-testable by
+  // node:test — McpFlowDiagram.tsx is JSX, which the test runner's
+  // `--experimental-strip-types` cannot parse. So /explore's "client" for
+  // this check is McpFlowDiagram.tsx plus the module it delegates to;
+  // QueryForm still does its own fetch inline.
   const clients = [
-    ['../../components/explore/McpFlowDiagram.tsx', '/explore’s live query'],
-    ['../../components/QueryForm.tsx', 'the home/ask query form'],
+    [
+      ['../../components/explore/McpFlowDiagram.tsx', '../../lib/offered-model.ts'],
+      '/explore’s live query',
+    ],
+    [
+      ['../../components/QueryForm.tsx'],
+      'the home/ask query form',
+    ],
   ] as const;
-  for (const [relative, what] of clients) {
-    const source = readFileSync(join(HERE, relative), 'utf8');
+  for (const [relatives, what] of clients) {
+    const sources = relatives.map((relative) => readFileSync(join(HERE, relative), 'utf8'));
+    const combined = sources.join('\n');
     // Read from the instance, not asserted here.
-    assert.ok(source.includes("'/api/models'"), `${what} reads the offered list`);
-    assert.ok(source.includes('parseModelsResponse'), `${what} uses the shared parser`);
+    assert.ok(combined.includes("'/api/models'"), `${what} reads the offered list`);
+    assert.ok(combined.includes('parseModelsResponse'), `${what} uses the shared parser`);
     // A model id sent on the wire may not be a literal in a client. Comment
-    // prose naming an id is fine and both files have some; what must not
+    // prose naming an id is fine and these files have some; what must not
     // appear is a literal in code, so the check is scoped to lines that are
     // not comments.
-    const codeLines = source
+    const codeLines = combined
       .split('\n')
       .filter((line) => !/^\s*(\/\/|\*|\/\*)/.test(line));
     for (const line of codeLines) {
--- a/src/components/QueryForm.tsx
+++ b/src/components/QueryForm.tsx
@@ -14,30 +14,12 @@ import {
   type QueryMode,
 } from '@/lib/query-presentation';
 import { parseModelsResponse, type Model } from '@/lib/model-list';
+import { MODELS_LOAD_ERROR } from '@/lib/streaming';
 import RateLimitBanner from './RateLimitBanner';
 
 // Re-exported for existing importers; the type itself lives in
 // src/lib/query-presentation.ts alongside the derivations that use it.
 export type { QueryMode };
-
-/**
- * Reader-facing copy for a `/api/models` load failure (#283) — covers both a
- * failed fetch/JSON parse and a 200 response whose body doesn't carry a
- * usable `models` array. Both are the same class of failure to the reader
- * (the model list didn't load), so both route through this one message
- * rather than a raw error or a silent render crash.
- *
- * It used to say "only the default model is available right now", which was
- * true while this form carried a hardcoded model id to fall back on. It no
- * longer does (website#30 P4): which models exist is instance configuration,
- * and a form that cannot read the list has nothing to substitute. So the
- * message now discloses a blocking state, because that is what it is — and
- * the send control is withdrawn rather than the picker being marked invalid,
- * per docs/design-principles.md Principle 3 and its corollary (a list that
- * failed to load is not a bad selection).
- */
-const MODELS_LOAD_ERROR =
-  "Couldn't load the list of AI models this site offers, so a query can't be sent right now. Refresh the page to try again.";
 
 /**
  * Sticky-per-session mode persistence (§10 Q7), on the shared
--- a/src/components/explore/McpFlowDiagram.tsx
+++ b/src/components/explore/McpFlowDiagram.tsx
@@ -11,7 +11,7 @@ import { useTraceReplay } from '@/hooks/useTraceReplay';
 import { useLiveTrace } from '@/hooks/useLiveTrace';
 import type { ReplayState } from '@/lib/bpmn/animation';
 import { traceEventsToProgressData } from '@/lib/bpmn/trace-progress';
-import { parseModelsResponse } from '@/lib/model-list';
+import { createOfferedModelResolver, type OfferedModelResolver } from '@/lib/offered-model';
 
 const DEFAULT_PORTAL = 'data.cityofnewyork.us';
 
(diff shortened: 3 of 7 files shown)


npstorey:
## Seat GO — P7 at `9ae7320`

**Head-bound GO for PR #317 at `9ae7320c989e82166398fd340bfc2973d4534d36`** (base `eadd284`, `rollback/pre-30-p7`). Owner: approve and merge this head. If the head moves, this GO lapses. This is the wave's last merge.

### Verified independently of the gate record (5412837975)

- **Numstat = the PR API:** 7 files, +303/−58 (the seven rows sum exactly); all under `src/`. One commit, signed off.
- **Rollup on the exact head:** `build / test / lint / typecheck` ✅ · `DCO` ✅ · `Vercel` (commit status) ✅ · `Vercel Preview Comments` ✅ — the ruleset's three required contexts among them. `standalone build + asset check` absent by its path filter (verified at P3; nothing here is under its paths).
- **Read in the code:**
  - **Criterion 1** — `src/lib/offered-model.ts` caches only a resolved, non-empty id; a network failure, a malformed body and an empty catalog all leave nothing cached, and `finally` clears the in-flight slot so the next call retries. The five tests are exactly the five behaviours (failure retries; malformed not cached; empty not cached; success cached; concurrent calls share one fetch).
  - **Criterion 2** — `modelsError` state in `McpFlowDiagram`, cleared on a later success; `TraceControls` disables Run and shows the shared `MODELS_LOAD_ERROR` in a `role="status"` element in place of the helper text; the query input is not marked invalid. `handleLiveStart` also guards an empty id and backs out of the fullscreen entry rather than opening it onto nothing.
  - **Criterion 3** — the corrected comment does not restate a description; it names the chain (`useLiveTrace` handles no `trace` event → `LiveResponsePanel` passes no `evidenceTrace` → `canPublish` requires it) and records what it corrects and why. The true defect is stated precisely: a model the instance does not offer was queried and billed, in a trace discarded on that surface.
  - **Criterion 4** — `compare-stream/route.ts` and `model-resolver.ts` are not in the diff.
  - **The shared string** moved to `streaming.ts` is byte-identical to the constant `QueryForm` dropped; the trimmed paragraph described wording that no longer exists — a trim, as recorded.
- **Sonnet 5 accounting, validated per G0-B:** 1004 + 5 = 1009 holds (the ordering test was updated — zero `test(` lines added or removed there); the seven-file list reconciles; the `react-hooks/refs` workaround (a lazy `useState` initializer) is disclosed with its reason. No slip this time. Two datapoints for the re-ask: one arithmetic slip on the doc-comment phase, none on the code phase.
- **Worktree isolation held** (shared checkout on `main`, clean).

### For the wave-close record

- The ledger states the series per phase — 2 / 4 / 4 / 4 / 0 / 3 / 0 — and its sum (**17** by the per-phase figures; the running "19" needs reconciling in the record), and states, as the ORCH already has, that P7's zero came with unusually good conditions.
- Filings to carry: the three pre-P6 preflight/app disagreements; the harness source registry spreading vendor titles and hostnames into the signed PROV graph; `CLAUDE.md`'s `# pass 872` (real: 1009 now); `/api/records`' tolerance of `body.model`; #312 / #314 / #315 open by design.
- Requirement 7's wording restated (the P1 preflight riders were chartered).

Ledger accepted as recorded: Sonnet 5 · one pass · 5/5 · 0 fix-on-top · 0 premises falsified.

— CAT PLAN (seat), 2026-08-25

Issue #325, opened 2026-08-26
Wave N6 — pre-deploy defect sweep: the answer path, and the diagnostics that should have caught it


npstorey:
## Scope

Six defects that must be fixed in the code an operator deploys, gathered into
one wave so the merges into `main` serialize through a single lane. Five were
found by running the app against a live Socrata portal; one is a build break
that has been on `main` since 2026-08-20.

They are grouped because three of them land in or beside the same 592-line
module, and because they are one story: **the answer path publishes things it
should not, and the diagnostics that would have caught it are themselves
broken.**

Members: #295 · #322 · #321 · #319 · #320 · #323. Related, different repo:
[socrata-mcp-server#58](https://github.com/npstorey/socrata-mcp-server/issues/58).

Out of scope, deliberately: #324 (the divergence triage — a design argument, not
a defect), and every product or layout question it raises.

## Why one wave and not six PRs

`src/lib/openrouter-streaming.ts` is the blast zone for **#322, #321 and #319**
simultaneously. Three concurrent branches on one 592-line module produce
conflicts that are resolved by whoever merges last, which is the worst possible
reviewer for that decision. The wave exists to serialize them.

## Phases

Each phase is one PR, merged on a head-bound GO plus the owner's approval.

**P1 — #295, the container build.** The `.dockerignore` allowlist gains the one
file the type-check needs, **and** a CI leg performs a real image build.

The CI half is not optional and not a nice-to-have: `.github/workflows/standalone.yml`
says so in its own header comment at line 7 — *"`docker build` — which no
workflow performs"*. That workflow runs `npm run build:standalone` **on the
host**, where `scripts/` is present. The image build has a `.dockerignore`-filtered
context where it is not. That gap is exactly why CI passed on the PR that broke
the image, and it will let the next one through too. P1 merges first; nothing
else in this wave can be container-verified until it does.

**P2 — #322, `resultSummary`.** Read the Socrata envelope
(`{"data": [...], "total_rows": N}`) in addition to the bare array. Note this
also switches on row-count narration in the live panels via `formatToolResult` —
two surfaces, both to be confirmed before merge. P2 comes before P3 because it
restores the diagnostic P3 was written to need.

**P3 — #321, replayed rejected calls.** `failed` + `failureKind` on the
tool-call record; rejected calls render as markdown rather than as an executable
cell; every surviving fetch cell `try`/`except`-wrapped.

**P4 — #319, the mid-plan answer.** A final answering turn with the output
[...]

PR #326, opened 2026-08-26
fix: stop advertising Socrata tools the model cannot call (#325 P6)

### LINKED ISSUE #323: System prompt advertises two Socrata tools that are never added to the tool list
## Summary

`CROSS_SOURCE_PREAMBLE` in `src/lib/mcp/socrata-skill.ts` tells the model it has
`get_data`, `search` and `fetch`. The routing registry's `SOCRATA_TOOLS` lists
all three, but only `get_data` is ever added to `mcpTools` — so the model is
told about two tools it has no way to call.

## Why it matters

A model that believes it has a `search` tool will plan around having one. The
failure is not an error message; it is a worse plan, chosen for a capability
that does not exist, and it is invisible in the output.

## Evidence

Read at `3464eb7`: the preamble names three tools (`socrata-skill.ts:483`); the tool
list is assembled with one.

## Suggested fix

Pick one, deliberately:

- **Add them** to `mcpTools` if the server implements them and they are wanted, or
- **Stop naming them** in the preamble.

The choice is a product decision about what the analysis loop should be able to
do; either resolution removes the mismatch. Whichever is chosen, a test that
asserts the preamble's named tools are a subset of the assembled tool list would
keep the two from drifting apart again.

npstorey:
Closes #323. Wave N6 phase P6.

## The defect

`CROSS_SOURCE_PREAMBLE` told the model it had three Socrata tools — `get_data`, `search` and `fetch`. Only `get_data` is ever added to `mcpTools`, so two of the three were uncallable.

This is not a visible failure. A model that believes it has a `search` tool plans around having one, and the resulting worse plan is invisible in the output — there is no error message to notice.

## The fix

The preamble now names `get_data` alone. `search` and `fetch` were **not** added to the tool list: `get_data` already covers catalog search, metadata, SoQL queries and metrics, so wiring them up would widen the analysis loop's decision surface with a redundant retrieval path on no measured demand. Adding them later is additive, and the new test keeps that honest whenever it happens.

`src/lib/mcp/registry.ts` is deliberately untouched. Its `SOCRATA_TOOLS` array is the *routing* registry — it builds `toolIndex` and `unconfiguredTools` so an instance with no Socrata endpoint refuses those names by naming the missing variable rather than routing to a default host. Removing names there would change refusal behaviour, which is a different concern.

## The guard

`src/lib/mcp/preamble-advertised-tools.test.ts` asserts that every tool name the preamble advertises is present in `mcpTools`. It derives the names from the preamble text itself rather than a copied list, so re-adding *any* uncallable name in future fails the test — not just these two. A structural assertion that exactly three `Tools:` lists are found means a fourth source forces a deliberate test update instead of silently passing through.

The test was demonstrated failing before it was accepted: temporarily restoring `search` to the preamble produces

```
not ok 1 - every tool name the cross-source preamble advertises exists in mcpTools
  error: 'Preamble advertises tool "search" but it is not in mcpTools ...'
```

## Diff

Two changed lines in `src/lib/mcp/socrata-skill.ts` — the Socrata bullet's tool list, and an `export` keyword so the test reads the live constant rather than a copy — plus one new test file.

```
 src/lib/mcp/preamble-advertised-tools.test.ts | 75 +++++++++++++++++++++++++++
 src/lib/mcp/socrata-skill.ts                  |  4 +-
```

## Checks

| Check | Result |
|---|---|
| `npm test` | 1010 tests, 1010 pass, 0 fail — +1 against the 1009 baseline, nothing moved |
| `npm run build` | Compiled successfully, route table printed |
| `npm run lint` | 4 problems, 0 errors, 4 warnings — the existing baseline, none introduced |
| `npm run check:compose-env` | PASS |

Every commit carries a DCO sign-off matching the author.

## Follow-up noted, not fixed here

`buildSystemPrompt(portal)` is the only production caller that assembles a prompt containing the real preamble, and it has no seam for substituting a registry — so its own wiring (which sources it activates, in what order, with which portal and preamble) is exercised only against the real network-calling registry. An optional defaulted registry parameter would close that gap. Additive, no caller changes required, and better filed as its own small test-infrastructure change than bundled here.

### DIFF
--- a/src/lib/mcp/preamble-advertised-tools.test.ts
+++ b/src/lib/mcp/preamble-advertised-tools.test.ts
@@ -11,68 +11,53 @@
  * in the output (civic-ai-tools-website#323 — the preamble told the model it
  * had `search` and `fetch` Socrata tools; only `get_data` was ever wired up).
  *
- * This test does not hardcode the expected tool list. It reads the preamble
- * source text directly, extracts every `Tools: a, b, c.` list via regex, and
- * checks each extracted name against the real `mcpTools` export. Re-adding
- * any uncallable name to the preamble in the future — not just `search` or
- * `fetch` specifically — fails this test.
+ * This test does not hardcode the expected tool list. It imports the real
+ * `CROSS_SOURCE_PREAMBLE` constant, extracts every `Tools: a, b, c.` list
+ * out of it via regex, and checks each extracted name against the real
+ * `mcpTools` export. Re-adding any uncallable name to the preamble in the
+ * future — not just `search` or `fetch` specifically — fails this test.
+ *
+ * Composing the full system prompt (`composeSkillPrompt` /
+ * `buildSystemPrompt`) was considered and rejected here: the property under
+ * test is a relation between the preamble text and `mcpTools`, and building
+ * intro + preamble + stub source bodies + outro around it doesn't strengthen
+ * that assertion — it only reintroduces a stray-"Tools:" hazard from
+ * whatever stub text stands in for the other sources. Importing the
+ * constant directly and asserting on it is the whole test.
  */
 
 import { test } from 'node:test';
 import assert from 'node:assert/strict';
-import fs from 'node:fs';
-import path from 'node:path';
-import { fileURLToPath } from 'node:url';
+import { CROSS_SOURCE_PREAMBLE } from './socrata-skill.ts';
 import { mcpTools } from './tools.ts';
-
-const HERE = fileURLToPath(import.meta.url);
-const MCP_ROOT = path.dirname(HERE);
-const SOCRATA_SKILL_SOURCE = path.join(MCP_ROOT, 'socrata-skill.ts');
-
-/**
- * Pulls the `CROSS_SOURCE_PREAMBLE` template literal's body out of the raw
- * source text. Scoping the search to just this declaration (rather than
- * scanning the whole file for "Tools:") keeps the extraction from picking up
- * an unrelated "Tools:" mention elsewhere in the module.
- */
-function extractCrossSourcePreamble(sourceText: string): string {
-  const match = sourceText.match(/const CROSS_SOURCE_PREAMBLE = `([\s\S]*?)`;/);
-  assert.ok(
-    match,
-    'Expected to find `const CROSS_SOURCE_PREAMBLE = `...`;` in socrata-skill.ts — has the declaration moved or been renamed?',
-  );
-  return match![1];
-}
 
 /**
  * Extracts every tool name out of every "Tools: a, b, c." list in the given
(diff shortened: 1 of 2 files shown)


npstorey:
## GO re-bound — head `a39718a86900389e59f5492015322ee7409fb914`

Supersedes the conditional GO at `5d7fc2e`
([5430004995](https://github.com/npstorey/civic-ai-tools-website/pull/326#issuecomment-5430004995)).
**The condition is discharged**: the gate P1 built has now evaluated P6.

### The diff was not re-read. It did not need to be — and that is a measurement, not trust.

The re-bind rests on the rebase being a faithful replay rather than a
re-authoring, so that is what was checked, at blob level:

| | `5d7fc2e` (validated) | `a39718a` (now) |
| --- | --- | --- |
| `src/lib/mcp/socrata-skill.ts` | `e507cf235a` | `e507cf235a` — **identical** |
| `src/lib/mcp/preamble-advertised-tools.test.ts` | `b8e8942ecc` | `b8e8942ecc` — **identical** |

Byte-identical content, so every judgement in the prior GO transfers without
re-reading a line. Two commits, both trailers preserved through the rebase
(`Signed-off-by` + `Co-Authored-By: Claude Sonnet 5`, no `Claude-Session`).

### Rebased onto the right thing, carrying nothing extra

- Merge-base is now `e053b5c` — P1's merge commit, current main.
- Three-dot against the **moved** base: +77 −2 across 2 files, reconciling with
  the PR API.
- `git diff --name-only e053b5c...a39718a` returns **exactly those two files** —
  nothing P1 brought is attributed to this branch. The three-dot was read
  precisely, as the phase said it was.

### Checks by name on this exact head

`DCO` · `Vercel` · `Vercel Preview Comments` · `build / test / lint / typecheck`
· **`container image build`** — all success. PR state moved `blocked` → `clean`.

That fifth one is the point of the rebase. When the ruleset gained
`container image build` as a required check, this PR became unmergeable because
it was cut before P1 and its context had no such job. It has now run against P6
and passed. The first merge to land after the gate exists is not the one that
skipped it.

`standalone build + asset check` is absent — and this time genuinely by path
filter: P6 touches only `src/lib/mcp/`, which matches none of that workflow's
five entries.

### Premise #8 is well made, and worth keeping

The earlier seat note that both legs were "absent by path filter" conflated two
absences that fail in **opposite directions**: `standalone` does not apply, which
is permanent and correct; `container image build` did not exist yet, which was a
gap and has now closed. Reading them as one thing is what would have let P6 merge
without the new gate ever looking at it. Recording the distinction rather than
the conclusion is the useful half.

---

## GO. Approve and merge.

On merge: tag `rollback/pre-325-p6-merged`, close #323, and wave 1 is complete —
both phases shipped, two keys on each, zero fix-on-top after a GO.

Issue #11, opened 2026-03-07
Add setup instructions for Claude/ChatGPT browser


npstorey:
Show how to enable MCP in Claude Desktop app. Note that ChatGPT doesn't support MCP natively. Could add to About page or create dedicated "Get Started" page. Link to modelcontextprotocol.io for official docs.

*Migrated from BACKLOG.md*

npstorey:
Done in Session 8. About page has Claude Desktop config (clone+build+stdio), Claude Code, Cursor cards, ChatGPT note, and MCP docs link.

PR #329, opened 2026-08-26
fix: every Data Commons cell is invalid Python (keyword argument repeated) (#325 P5)

npstorey:
Closes #320. Wave N6 phase P5.

## The defect

`pyKwargs` appends every argument the caller did not enumerate. The Data Commons renderer emits `variable_dcid` and `place_dcid` by hand **and** lets `pyKwargs` append them again, so both land twice and the cell is a `SyntaxError: keyword argument repeated`.

Not one failing notebook — **every** notebook containing a Data Commons fetch, since the renderer was written.

The Socrata renderer escaped only by accident: it hand-writes `portal` and `dataset_id`, and those two happened to be in a hard-coded skip list inside `pyKwargs`. That list was a stand-in for the mechanism this change actually builds — the renderer and `pyKwargs` each believed they owned the argument list, and one pair of keys happened to be covered while the other was not.

## The fix

`pyKwargs` gains a `handledKeys` parameter that seeds its `seen` set. Each caller declares what it has already rendered:

- `renderSocrataQueryCell` → `['portal', 'dataset_id', 'type']`
- `renderDataCommonsObsCell` → `['variable_dcid', 'place_dcid']`

The hard-coded skip list is **removed**, not widened — the caller now owns the declaration, which is the point. `type` is a call-routing discriminator that selects which renderer runs and is never itself a rendered kwarg; it is declared by the one caller that carries it rather than given a second "never a kwarg anywhere" concept.

"Append everything the caller did not enumerate" is preserved for every renderer, so a field a renderer does not know about yet is still emitted rather than silently dropped.

## The guard, and why it is not `ast.parse`

Every generated cell, across every renderer, is now compiled as Python by a test that shells out via `spawnSync`.

The obvious implementation — `ast.parse()` — **cannot detect this defect**:

```
$ python3 -c "import ast; ast.parse('f(x=1, x=2)')"     # no error
$ python3 -c "compile('f(x=1, x=2)', '<c>', 'exec')"    # SyntaxError: keyword argument repeated: x
```

Duplicate-keyword detection happens during `compile()`, not during grammar-level parsing. A parse test written the obvious way would have passed against the broken renderer and been indistinguishable from a working guard. The test uses `compile()` — a strict superset of `ast.parse()`'s syntax checking — and says why in its own docstring.

That distinction was found by measurement, not by reading: the test was demonstrated **red** against the pre-fix renderer before it was accepted.

```
not ok 1 - every renderer emits code cells that parse as Python
    SyntaxError: keyword argument repeated: variable_dcid
not ok 2 - data commons: date and child_place_type each render exactly once when present
not ok 3 - data commons: variable_dcid and place_dcid render exactly once even without date/child_place_type
ok 4 - socrata: rendered cell output is unchanged        <- Socrata never had this bug
```

The Socrata case staying green in that run is itself evidence the fix is targeted: it proves the test distinguishes the broken renderer from the working one rather than failing everything indiscriminately.

## Checks

| Check | Result |
|---|---|
| `npm test` | 1014 tests, 1014 pass, 0 fail — 1010 baseline + 4 new, none moved or removed |
| `npm run build` | Compiled successfully, route table printed |
| `npm run lint` | 4 problems, 0 errors, 4 warnings — existing baseline |
| `npm run check:compose-env` | PASS |

Socrata's rendered cell is asserted unchanged by exact string equality, not merely by absence of a visible diff.

## Coordination note

This file is also P3's blast zone (#321 adds failure fields to `PhaseAToolCall` and changes rejected-call rendering). The diff is kept tight and local to the kwargs mechanism so P3 composes onto it rather than conflicting. P5 merges before P3 is cut.

### DIFF
--- a/src/lib/notebook-author/tool-to-cell.test.ts
+++ b/src/lib/notebook-author/tool-to-cell.test.ts
@@ -1,0 +1,175 @@
+// Renderer syntax + kwargs-collision tests (Wave N6 P5, #320).
+//
+// pyKwargs appends every arg a renderer did not enumerate by hand. The Data
+// Commons renderer hand-writes variable_dcid and place_dcid, then also
+// passes call.args through pyKwargs — so both landed twice and every
+// generated Data Commons cell was a Python SyntaxError: keyword argument
+// repeated. This is the first test file for tool-to-cell.ts. It:
+//
+//   - parses every generated code cell as Python via a real python3
+//     subprocess, across every cell-producing renderer in the file — not
+//     just the Data Commons one, so a repeat-kwargs regression anywhere
+//     here trips it;
+//   - pins that a Data Commons call carrying date/child_place_type renders
+//     each kwarg exactly once;
+//   - pins the Socrata renderer's output as byte-for-byte unchanged.
+//
+// Run with: npm test
+
+import { test } from 'node:test';
+import assert from 'node:assert/strict';
+import { spawnSync } from 'node:child_process';
+import type { NotebookCell } from './cells.ts';
+import { renderFetchToolCell, type PhaseAToolCall } from './tool-to-cell.ts';
+
+const CTX = { dataFrameIndex: 1, defaultPortal: 'data.cityofnewyork.us' };
+
+/**
+ * Parse a Jupyter code cell's source as Python via a real `python3`
+ * subprocess — the same syntax check `nbconvert`/the sandbox execution
+ * path would trip over first.
+ *
+ * Uses `compile()`, not bare `ast.parse()`: measured locally, `ast.parse()`
+ * accepts a repeated keyword argument (e.g. `f(x=1, x=2)`) without error —
+ * that check runs during compilation, not grammar parsing — so it would
+ * pass on the pre-fix #320 renderer output and never demonstrate the
+ * failure this test exists to catch. `compile(..., 'exec')` is a superset
+ * of `ast.parse()`'s syntax checking and also raises
+ * `SyntaxError: keyword argument repeated` for that case.
+ */
+function assertParsesAsPython(cell: NotebookCell, label: string): void {
+  assert.equal(cell.cell_type, 'code', `${label}: expected a code cell`);
+  const source = cell.source.join('');
+  const result = spawnSync(
+    'python3',
+    ['-c', "import sys; compile(sys.stdin.read(), '<cell>', 'exec')"],
+    { input: source, encoding: 'utf-8' },
+  );
+  assert.equal(
+    result.status,
+    0,
+    `${label}: generated cell is not valid Python (exit ${result.status})\n` +
+      `--- source ---\n${source}\n--- stderr ---\n${result.stderr}`,
+  );
+}
+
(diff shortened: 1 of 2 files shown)


npstorey:
## GO — head-bound at `9fc8c6b`

Validated independently, fetch-only, no checkout.

| | |
| --- | --- |
| Merge-base | `02ca11a` — current main |
| Three-dot | +197 −5 across 2 files — reconciles with the PR API |
| Files | `tool-to-cell.ts`, `tool-to-cell.test.ts` — exactly the declared blast zone |
| Commit | 1, sign-off present, `Co-Authored-By: Claude Sonnet 5`, no `Claude-Session` |
| Checks by name on this head | `DCO` · `Vercel` · `Vercel Preview Comments` · `build / test / lint / typecheck` · `container image build` — all success |

### Ledger #11 is confirmed, and the criterion was the seat's as much as yours

Reproduced independently on python 3.9.6:

```
ast.parse("f(x=1, x=2)")                  -> no error
compile("f(x=1, x=2)", "<t>", "exec")     -> SyntaxError: keyword argument repeated: x
```

And on the real shape — `fetch_data_commons(variable_dcid=…, place_dcid=…, variable_dcid=…)`
— `ast.parse` is silent while `compile` raises `keyword argument repeated:
variable_dcid`. A test built to the criterion as written would have passed green
on the broken renderer.

**The seat wrote `python3 (import ast)` into the relay that carried that
criterion**, so this is not solely the ORCH's to own. It is the #295 defect class
— a check running in a mode where it cannot see what it was built to catch —
reproduced inside the criterion written to prevent it. Twice in one wave, from
two different authors, which is the useful part of the finding: it is not a
lapse, it is a shape people reliably fall into.

The phase did the right thing with it: `spawnSync('python3', ['-c', "import sys;
compile(sys.stdin.read(), '<cell>', 'exec')"])`, with the reasoning written into
the test file rather than left in a gate record — including that `compile` is a
superset of `ast.parse`'s checking. A future reader who "simplifies" it back to
`ast.parse` now has to argue past a comment.

### The fix is better than the issue proposed

The issue said callers should declare what they rendered by hand. The phase went
one step further: `pyKwargs` gains a `handledKeys` parameter and the hardcoded
`type` / `portal` / `dataset_id` skip-list is **deleted**. That list only knew
about Socrata, which is precisely why the Data Commons renderer's hand-written
keys were duplicated. Ownership moves to the caller that actually knows.

The red-run evidence is also stronger than asked for: with the fix reverted, the
Data Commons case fails **while Socrata stays green**. That proves the test
discriminates rather than failing indiscriminately — the difference between a
test and an assertion that everything is fine.

---

## GO.

**Hard constraint, restated because it is the reason wave 2 is a pair:** P5 must
merge **before P3 is cut**. They share `src/lib/notebook-author/tool-to-cell.ts`
— P5 at `pyKwargs` and the Data Commons renderer, P3 at `PhaseAToolCall` and the
same renderer family. That collision is ledger #1, the anchor's false
disjointness claim, and it must not be re-derived at the wrong moment.

Issue #338, opened 2026-08-27
Replay route carries the mid-plan-answer defect and feeds a signed consistency attestation


npstorey:
## Summary

`src/app/api/evidence/[slug]/replay/route.ts` runs its own copy of the tool-calling loop and carries the defect just fixed elsewhere — twice — while feeding a machine-generated consistency attestation against a signed record.

This is live on the reference instance today. It is not a deployment question: that instance signs what it publishes, and this path contributes to what is signed.

## The three defects, measured at `1ddd61a`

**1. A mid-plan narration becomes the replay's answer — the defect from the answer-path fix, verbatim.**

`:218` and `:238`:

```ts
let output = response.choices[0]?.message?.content || '';
if (!output && response.choices[0]?.message?.tool_calls) { … }
```

Any assistant turn carrying content is taken as the answer, and the guard is keyed on content being *absent* rather than on the content satisfying anything. A turn that announces its next query is published as the replay's output.

**2. Raw error text is fed to the model.** `:195`:

```ts
content: `Error: ${error instanceof Error ? error.message : 'Unknown error'}`
```

`describeToolFailureForLlm` exists for exactly this and is not used anywhere in this file. Status codes, host names and stack fragments reach the model, which is the condition the anti-hallucination guard was built to remove.

**3. The tool-call record carries no failure or result information.** Calls are recorded as `{name, args}` only — no `resultSummary`, no `failed`/`failureKind`. The file also carries its own private bare-array `truncateToolResult` (`:27-42`), with the same envelope defect tracked in #331.

## Why the severity

`src/components/evidence/AttestationDialog.tsx:233` calls this route N times and derives consistency metrics from the results. Those metrics are submitted as an **attestation against a published record**.

So a replay run that announces a query instead of answering one becomes an input to a numerical claim about how consistently that record reproduces. The reader of the attestation sees a consistency score; nothing distinguishes a run that answered from a run that said it was about to.

## Scope note

This route is one of several parallel implementations of the model-calling loop — see the companion proposal for a census of the class. Fixing this file alone would repair one of at least three loop-exit implementations, which is the error pattern this wave's cold read identified. It is filed separately because of the attestation chain, not because it should be fixed in isolation.

npstorey:
The companion census this issue's scope note refers to is now #345 (Wave N7 — the model-calling loop as a class, measured at 1ddd61a). This route is P3 there, on a shared loop core rather than a fix in place. The third copy, openrouter.ts, is #344.

PR #346, opened 2026-08-27
docs: wave N6 close — retro, corrected CI facts, and the class-scope rule (#325)

npstorey:
Closeout for wave N6 (anchor #325). Docs only — no source changes.

**Three CLAUDE.md claims this wave falsified**, corrected: the commands table's test count (872 → 1044), the description of `ci.yml` as one job when it has had two required jobs since #295, and the `check:standalone` line, which now also runs inside the container job — with the distinction that matters stated, namely that host and container are different environments and a green run on one is not evidence about the other.

**One rule added**, in the section's existing incident-comment form: *the model-calling loop has several implementations — fixing one is not fixing it.* That was the wave's headline finding. Three phases each fixed one instance of a defect that lived in a class, each honouring its blast zone exactly: #319 was fixed in `openrouter-streaming.ts` while `openrouter.ts:120` kept the literal pre-fix condition and `replay/route.ts` kept it twice while feeding a signed attestation (#338). The same shape hit the preamble (#323, reopened) and the notebook path (`notebook.ts:123`).

**A retrospective entry** recording the six phases, the falsified-premise ledger and its attribution (four the anchor's, five the seat's, ten the ORCH's, one a phase's), and the two acceptance criteria that close honestly qualified rather than ticked.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -29,7 +29,7 @@ remote) — don't design an owner-run leg around the assumption that it doesn't.
 
 | Command | Healthy output |
 |---|---|
-| `npm test` | `# pass 872` / `# fail 0` (`node --test` TAP summary) |
+| `npm test` | `# pass 1044` / `# fail 0` (`node --test` TAP summary) |
 | `npm run build` | `✓ Compiled successfully`, then the Route (app) table |
 | `npm run typecheck` | no output, exit 0 — **run after `npm run build`** (tsconfig includes `.next/types`, which the build emits) |
 | `npm run lint` | `✖ 4 problems (0 errors, 4 warnings)` — warnings are the baseline; **zero errors** is the gate |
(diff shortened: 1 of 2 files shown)


npstorey:
Seat read of `c4afcac` (three-dot against `main` `1ddd61a`: +67/−5, two files, one signed-off commit, sig G; all four required checks green on this exact head — `build / test / lint / typecheck`, `container image build`, DCO, Vercel).

**Every CLAUDE.md correction holds against the tree.** 1044 is the count from the runner's log at `1ddd61a`; `ci.yml` has two jobs and the ruleset on `main` requires both; `check:standalone` reaches the container job through `Dockerfile:63` (`npm run build:standalone`). The new rule's list of five other call sites matches the census at #345.

**Two sentences in `docs/RETROSPECTIVE.md` need a fix on top before this merges** — both are counts a reader will quote, and one already has been:

1. *"Six files call the model outside the one that was fixed."* — Five. Six in total, counting `openrouter-streaming.ts`; the CLAUDE.md rule in this same PR lists the five correctly. (#345's census: six non-test files call `chat.completions.create`.)
2. *"…left `openrouter.ts:120`, which still carries the literal pre-fix condition, plus `replay` and `evaluate`."* — `evaluate/route.ts:109` is a single rubric call with no tool loop and no loop-exit condition (#345's single-call table). The copies P4 left are `openrouter.ts` and `replay/route.ts`; drop "and `evaluate`".

One commit, docs only. **GO is conditional on that fix and re-binds to the new head**; a moved head voids this read, so post the SHA when it's up. After the merge: this checkout is sitting on `docs/wave-n6-closeout` — return it to `main` (`git checkout main && git pull --ff-only`) before closing, since ORCH 345 has booted into the same tree and is holding all git writes until it hears the tree is back on `main`.

npstorey:
**GO — head-bound to `1bdb818`.** Re-read three-dot against `main` `1ddd61a`: +67/−5 over the same two files; the delta from `c4afcac` is the one retrospective sentence, now correct (five outside, six in total; `evaluate` dropped from the loop list). Both commits signed-off, sig G. Required checks on this exact head: `build / test / lint / typecheck`, `container image build`, DCO, Vercel — all SUCCESS; `mergeStateStatus: CLEAN`. The anchor's close record carries the same correction. Owner's key: merge commit. A moved head voids this GO.

Issue #349, opened 2026-08-27
Tool-call arguments are parsed without a guard in all three loops; a malformed argument aborts the run


npstorey:
## Summary

All three tool-calling loops parse the endpoint's tool-call arguments with a
bare `JSON.parse` and no guard. Malformed arguments throw out of the loop, and
each caller turns that into a different kind of failure — on the replay path, a
500 in the middle of an N-run consistency test.

Measured at `7357f27`:

```
src/lib/openrouter-streaming.ts:421          const args = JSON.parse(toolCall.function.arguments);
src/app/api/evidence/[slug]/replay/route.ts:171   const args = JSON.parse(toolCall.function.arguments);
src/lib/openrouter.ts:85                     const args = JSON.parse(toolCall.function.arguments);
```

Every one of them sits **outside** the `try` that wraps tool execution, so the
existing failure handling — `describeToolFailureForLlm` on the streaming path,
the `Error:` string on the other two — never sees it.

## What each caller does with the throw

- **`openrouter-streaming.ts`** — escapes to the outer catch and reaches the
  reader through `reportStreamFailure` → `friendlyStreamError`. Sanitized, but
  the run is over: a single malformed argument ends a query that may already
  have completed several successful tool calls.
- **`replay/route.ts`** — escapes to the route catch (`:254-275`).
  `classifyModelError` does not classify a `SyntaxError`, so it falls through to
  `NextResponse.json({ error: message }, { status: 500 })`. The client
  (`AttestationDialog.tsx:241-243`) throws on the first non-ok response, so one
  malformed argument aborts the whole consistency run — after N−1 replays have
  already been paid for.
- **`openrouter.ts`** — escapes to `compare/route.ts`'s handler.

## Why it is filed as a class

This is the same shape as #338, #344 and #347: one defect, three
implementations, no single place to fix it. It is filed now because Wave N7
(#345) is consolidating these three loops into one core, and the moment the
parse exists in exactly one place is the cheap moment to guard it — a malformed
argument should be recorded as a failed tool call (`failed` + `failureKind`,
#321's shape) and described to the model through
`describeToolFailureForLlm`, exactly as an execution failure already is.

Not in that wave's charter. Filed so the census of this family stays complete
rather than rediscovering it next time.

## Provenance

Found during Wave N7's P1 design phase while reading all three loops against
each other; recorded here rather than fixed, per that phase's blast zone (no
source files).

npstorey:
[...]

Issue #348, opened 2026-08-27
The adversarial-eval rubric call exists twice: runAdversarialEval and evaluate/route.ts


npstorey:
## Summary

The adversarial-evaluation rubric call exists twice: once in
`src/lib/evidence/adversarial-eval.ts:74`, once copied into
`src/app/api/evidence/[slug]/evaluate/route.ts:109`. Same rubric, same prompt
builder, same `max_tokens`; they differ only in where the key comes from and how
errors are handled.

Neither copy carries a tool-calling loop, so neither is in the loop family Wave
N7 (#345) is scoped to. Harm today is nil. This is filed so the next census does
not rediscover it as a finding.

## Measured at `7357f27`

| | `runAdversarialEval` | `evaluate/route.ts` |
|---|---|---|
| Call site | `adversarial-eval.ts:74` | `:109` |
| Key | platform key via `getModelClient()` | caller-supplied, `createModelClient({apiKey})` |
| Reached from | `[slug]/publish/route.ts` at publication | `AttestationDialog.tsx:270` |
| Signed | directly — `emitEvaluationAttestation` signs an `attestation/evaluates/v1` node, RFC3161 + Rekor | via the client: result → `POST /attestations` → signed |
| Tests | the core has them; the runner does not | none |

Two members, and the divergence is real but small: key sourcing and error
handling are the only axes on which the copies differ.

## One thing worth correcting while here

`evaluate/route.ts:120` extracts the model's text with a bare
`response.choices[0]?.message?.content || ''`. That is the same extraction shape
as the loop-family defect, but it is not the same defect: the result is handed
straight to `parseEvaluationResponse` (`:122`), and a narration that is not a
well-formed rubric response fails that parse and returns 502 rather than being
published. The structural check is what makes it safe, and it is worth keeping
in mind if the parse is ever loosened.

## Suggested direction

Have the route call `runAdversarialEval` with a caller-supplied key, so key
sourcing becomes a parameter of one function rather than the reason for a second
copy. Not urgent; it belongs with whatever decision is taken about tightening
`/api/compare*`, which the route's own comment at `:91` already notes.

## Provenance

Registered in the Wave N7 census (#345) and ruled out of that wave at its G0
gate (D4 = A, 2026-08-27) so the wave stays scoped to the loop family whose
fix-shape is under test.

npstorey:
[...]

Issue #347, opened 2026-08-27
Replay's token-budget break duplicates an assistant turn and re-answers its tool calls


npstorey:
## Summary

On the token-budget path, the replay route sends the model a transcript that
duplicates an assistant turn and answers the same `tool_call_id` twice. The
request is malformed, and the run it produces is an input to a signed
consistency attestation.

This is a fourth defect in `src/app/api/evidence/[slug]/replay/route.ts`,
alongside the three in #338. It was not in the Wave N7 census; it was found
reading the file against the streaming implementation, which guards this exact
case and has a regression test for it.

## Measured at `7357f27`

The loop pushes each assistant turn before executing its tools (`:166`):

```ts
const assistantMessage = response.choices[0].message;
messages.push(assistantMessage);
```

The token-budget check then breaks out of the loop **before** the next request
is made (`:201`):

```ts
// Token budget check
if (cumulativeTokens >= 200_000) break;
```

Because that `break` precedes the `create` call at `:203`, `response` still
holds the turn whose message was already pushed at `:166`, and whose tool calls
were already answered with real results at `:186-196`.

The terminal block then treats it as unanswered (`:221-229`):

```ts
if (!output && response.choices[0]?.message?.tool_calls) {
  messages.push(response.choices[0].message);          // second push of the same turn
  for (const tc of response.choices[0].message.tool_calls) {
    messages.push({
      role: 'tool',
      tool_call_id: tc.id,                             // already answered above
      content: 'Limit reached. Please summarize based on data collected.',
    });
  }
```

So on the token-budget path the request carries the assistant turn twice and two
`tool` messages for every `tool_call_id` — one with the real result, one with the
placeholder.

The max-iterations path is **not** affected: when the `while` condition fails on
`iterations >= MAX_ITERATIONS`, `response` has been refreshed at `:203` and its
message was never pushed, so the terminal block's push is correct.

## The same case, handled, one file away

`src/lib/openrouter-streaming.ts` carries a flag for exactly this, with the
reasoning written down at `:398-405`:

> each assistant turn before executing its tools, so the token-limit break
> below leaves the loop with the final message already in the transcript,
> its tool calls already answered with real results. The terminal block
> must not push it a second time: that would duplicate the assistant turn
> and answer the same `tool_call_id` twice, which is a malformed request.

`lastMessageAlreadyInTranscript` is set at `:413`, cleared only past a fresh
request at `:579`, and gates both the re-push (`:630`) and the placeholder
[...]

Issue #344, opened 2026-08-27
openrouter.ts carries the original tool loop: pre-#334 exit condition, raw error text to the model, no truncation, name/args-only records — reachable via /api/compare


npstorey:
## Summary

`src/lib/openrouter.ts` is the **original** tool-calling loop — the one `openrouter-streaming.ts` was forked from on 2026-01-22 — and it has not received any of the fixes that landed on the fork. It is reachable today through `POST /api/compare`, which is the query the operator docs use as the smoke test for a new instance.

Companion to #338 (the replay route's copy). This one is filed separately because it has no issue of its own; #338's scope note already says neither should be fixed in isolation.

## Measured at `1ddd61a`

**1. The pre-fix loop exit, verbatim.** `:120`:

```ts
if (!lastMessage?.content && (iterations >= maxIterations || lastMessage?.tool_calls)) {
```

`git blame` dates this line to `cac7baa` (2026-01-14). The streaming file's condition at #334's base (`1ddd61a^1:495`) was this line plus `tokenLimitExceeded`; #334 replaced it because a final turn carrying prose was shipped as the answer even when that prose announced a query the model never ran (#319). Here the same guard is keyed on content being *absent*, so the same narration is returned as `content`.

**2. Raw error text reaches the model.** `:99`:

```ts
content: `Error executing tool: ${error instanceof Error ? error.message : 'Unknown error'}`,
```

`describeToolFailureForLlm` (`src/lib/streaming.ts:332`) is not imported. `CLAUDE.md`'s rule — LLM-facing tool-failure strings go through it — is stated for streaming paths; this is the non-streaming one.

**3. No truncation at all.** `:93` pushes the tool result as returned. The streaming file bounds it at `MAX_TOOL_RESULT_CHARS` (with the envelope defect in #331); this file has no bound, so a large Socrata result is handed to the next turn whole.

**4. The tool-call record carries no failure or result information.** `CompletionResult.tools_called` (`:9-12`) is `{name, args}` only — the shape #321 changed on the streaming path because a failed call was indistinguishable from a zero-row one.

**5. The final-answer request is the wording #334 replaced.** `:141` still reads *"Based on all the data you have collected from the tool calls above, please provide a comprehensive answer to my original question. Summarize the key findings."* — the sentence #334's own comment quotes as the previous wording.

Also: `maxIterations` is 10 here, 20 in the other two loops; `max_tokens` is 2000, not 4000; there is no token budget.

## Reachability

[...]

PR #350, opened 2026-08-27
feat: one tool-calling loop, and the streaming caller on it (#345 P2)

### LINKED ISSUE #331: truncateToolResult feeds the model malformed JSON for any oversized Socrata envelope
## Summary

`truncateToolResult` recognises only a bare JSON array. A Socrata envelope — `{"data": [...], "total_rows": N, ...}` — is not an array, so an oversized result falls through to raw character truncation and is cut mid-record. The string handed back to the model is **not valid JSON**.

This is the same premise that made `resultSummary` always-null for Socrata (#322, fixed in #330). That fix corrected the diagnostic path; this is the same wrong assumption two functions away, on the path that feeds the model.

## Evidence

Measured 2026-08-26 during #322's phase, with a 2,000-row Socrata-shaped envelope run through the actual `truncateToolResult` logic:

```
envelope length: 296028
output length:    50042
last 200 chars:  ...,"borough":"BROOKLYN","incident_zip":"11201"},{"unique_key":"100337",
                 "created_date":"2026-01-01T00:00:00.000","complaint_type":"Noise - Residential","borough":"B
output is valid JSON: false
  Bad control character in string literal in JSON at position 50000
```

The bare-array branch does the right thing — it slices whole rows and appends `[Truncated: showing N of M rows]`. The envelope never reaches it.

## Why it matters

The truncated string becomes the `tool` message content fed back into the next model turn. A model handed malformed JSON gets no signal that truncation happened — there is no `[Truncated: ...]` marker on this path — and the final record is built on whatever it makes of a fragment ending mid-string.

`MAX_TOOL_RESULT_CHARS` exists to bound input token growth; that bound is still enforced. The defect is in what the bounded output *is*.

## Location

`src/lib/openrouter-streaming.ts`, `truncateToolResult` — the gate is `if (Array.isArray(parsed) && parsed.length > 0)`.

## Suggested fix

Extract the `data` array the way the summary path now does, truncate by row count as the bare-array branch already does, and re-emit the envelope framing closed correctly so the result stays parseable — keeping the `[Truncated: showing N of M rows]` marker so the model is told.

## Sequencing note

Deliberately not fixed in #330. It is a context-management change landing on the same function region that the final-answering-turn work (#319) rewrites for token limits, and those changes were serialized on purpose to keep two branches off one module. Best done with #319 or immediately after it, by whoever holds that file next.

npstorey:
# P2 — the shared loop core, and the streaming caller on it

Wave N7 (#345) phase 2. Branch `sprint-345/p2-shared-loop-core`, cut from `7357f27`.
Two commits: `059af3b` (the core and the streaming caller on it) and `9edd841` (one
extension to #349's guard, itemised below). Model: **Opus 5** (owner ruling D5 = C).
Closes #331, #343, #349.

## Diff stat and blast zone

```
 src/lib/model-identity.test.ts                 |  54 +-
 src/lib/model-loop/model-call-registry.test.ts | 193 ++++++
 src/lib/model-loop/run-tool-loop.test.ts       | 472 ++++++++++++++
 src/lib/model-loop/run-tool-loop.ts            | 848 +++++++++++++++++++++++++
 src/lib/model-loop/test-harness.ts             | 127 ++++
 src/lib/openrouter-streaming.test.ts           | 115 +---
 src/lib/openrouter-streaming.ts                | 704 ++++-----------------
 7 files changed, 1827 insertions(+), 686 deletions(-)
```

**Every file is inside the declared zone.** `src/lib/model-loop/**` (new, including the
registry test's own file), `openrouter-streaming.ts` and its test, and
`model-identity.test.ts` for the pre-authorized D6 drift-guard replacement only. Nothing
else in the tree is touched — no route, no consumer, no doc, no config. **No out-of-zone
edit was needed, so none is itemised.**

## What landed

`src/lib/model-loop/run-tool-loop.ts` exports `runToolLoop` to G1's signature. Carried in:
#334's three-way exit (`abortedMidPlan || answeredNothing || announcesUnrunWork`),
`describeToolFailureForLlm` on every tool failure, failure-bearing tool records
(`failed`/`failureKind`, #321's shape), and the `lastMessageAlreadyInTranscript`
discipline. `queryWithMcpStreaming` / `queryWithoutMcpStreaming` keep their external
signatures; `announcesUnrunWork`, `responseModelAttributes` and `TraceContext` are
re-exported from `openrouter-streaming.ts`, so no importer moves.

What is core behaviour and what is a parameter follows G1's ruling: anything a caller
could opt out of that would reintroduce a class defect has no switch (the exit condition,
the failure vocabulary, the error-to-model path, transcript well-formedness, the
truncation, the single contract statement); what a caller is given (client, model, tools,
budgets, tool transport) and what it does with the result (progress rendering, streaming,
error surfacing) are parameters. `PanelType` does not appear anywhere under
`src/lib/model-loop/`, and every import there is relative.

## Streaming behaviour is unchanged — measured, not asserted

The base implementation was materialized from `7357f27` beside the new one and both were
driven through the same scripted endpoint, the same tool transport and a **deterministic
trace clock and id source**. Compared: every callback invocation in order, every request
body the endpoint received, and the full finalized OTel trace.

```
  ok   pass-through answer — callbacks / requests / trace
  ok   answering turn after a narration — callbacks / requests / trace
  ok   token budget exhausted — callbacks / requests / trace
  ok   failing tool call — callbacks / requests / trace
  ok   two tool calls in one turn — callbacks / requests / trace
  ok   oversized envelope — callbacks, trace
  DIFF oversized envelope — requests
       base: …"unique_key":"11\n[Truncated: result was 84918 characters]"}],"tools":[]…
       core: …"unique_key":"1188","borough":"BROOKLYN"},{"unique_key":"1189"…
  ok   portal injected inside the tool closure — callbacks / requests / trace
```

20 of 21 comparison groups byte-identical; the one divergence is #331's fix, in the tool
message the model receives. Wall-clock measurements (`duration_ms`, span timestamps) were
elided — they jitter run to run and are not behaviour.

### 1. The core exists; streaming runs on it; consumers are untouched — PASS

```
$ git diff 7357f27...HEAD -- src/app/api/compare-stream/route.ts src/app/api/query-notebook/route.ts | wc -l
0
$ git diff 7357f27 --stat -- src/app/api/
(no output)
```

### 2. Full suite green, nothing removed — PASS

`# tests 1065 · # pass 1065 · # fail 0` (baseline 1044 + 21 new). Test names compared, not
just counts — every `test('…')` name in every test file at `7357f27` against HEAD:

```
$ comm -23 /tmp/all-base.txt /tmp/all-head.txt     # present at base, absent at HEAD
(empty)
1044 names at 7357f27 · 1065 at HEAD
```

### 3. #334's exit logic is load-bearing in the core — SHOWN RED, THEN GREEN

*(This demonstration and the next were run at `059af3b`, before the second commit added
one test — hence 1064 rather than 1065 in their counts.)*

The core's condition was mutated to the pre-#334 form
(`!lastMessage?.content && (iterations >= maxIterations || lastMessage?.tool_calls)`) and
the full suite run:

```
not ok 823 - #334: a final turn that announces a query is not returned as the answer
not ok 826 - a blocking final turn answers without streaming — the shape a route caller uses
not ok 895 - #319: a tool_calls-free message that ANNOUNCES the next query is not published as the answer
not ok 897 - #319: at most ONE answering turn — a second unsatisfying answer is not re-asked forever
not ok 898 - #319: a token-limit-exceeded run with content asks for a summary instead of shipping that content
not ok 899 - #319: the answering turn omits tools, restates the contract, and keeps the transcript well-formed
not ok 900 - #319: the token-limit answering turn does not duplicate the assistant turn or re-answer its tool call
not ok 901 - #345 P2: token_limit_exceeded is present on the answering-turn completion and ABSENT on the pass-through one
# tests 1064
# pass 1056
# fail 8
```

Mutation reverted:

```
# tests 1064
# pass 1064
# fail 0
```

All five #319 cases go red on the mutation, plus the two core-level cases and the F-1 pin.

### 4. The registry test exists and can fail — SHOWN RED, THEN GREEN

`src/lib/model-loop/model-call-registry.test.ts`. A throwaway
`src/lib/throwaway-second-loop.ts` calling `chat.completions.create` with `tools` was
added:

```
  error: 'src/lib/throwaway-second-loop.ts calls chat.completions.create and is not on the registry in
          src/lib/model-loop/model-call-registry.test.ts. Add it with one line saying why it is a
          separate call site — or, better, call runToolLoop.'
  error: 'src/lib/throwaway-second-loop.ts calls chat.completions.create with tools — a second
          tool-calling loop. This is the shape wave #345 exists to end: call runToolLoop instead.'
# tests 1064 · # pass 1062 · # fail 2
```

Removed:

```
# tests 1064 · # pass 1064 · # fail 0
```

The allowlist at this head, each entry with its reason:

| File | Why |
|---|---|
| `src/lib/model-loop/run-tool-loop.ts` | The shared core |
| `src/lib/openrouter-streaming.ts` | `queryWithoutMcpStreaming`: the no-tools A-side. One turn, no loop, not loop-class |
| `src/app/api/evidence/[slug]/replay/route.ts` | Its own loop, **until P3** (#338, #347) |
| `src/lib/openrouter.ts` | Its own loop, **until P4** (#344) |
| `src/lib/evidence/adversarial-eval.ts` | One rubric call, no loop. Out of the wave by D4 (#348) |
| `src/app/api/evidence/[slug]/evaluate/route.ts` | The second copy of that rubric call. Out by D4 (#348) |
| `src/app/api/evidence/generate-summary/route.ts` | One summary call, no loop. Out of the wave |

The test is **two assertions on one instrument**, per G0 item 1: the registry (who calls
the model at all) and the narrower tool-loop count (which calls pass `tools` — three at
this head, one when the wave closes). It parses each call's argument block rather than
grepping, so `model-client.ts:458`'s doc comment is not counted and `openrouter.ts`'s
`tools,` shorthand is. Both directions are asserted: an unlisted caller fails, **and a
stale allowlist entry fails** — the list cannot rot into a false statement about the tree.

### 5. #331 and #343 close — PASS, each shown red

**#331.** A 2,000-row envelope (296,028 characters — the same size as the issue's own
reproduction) is driven through the core and the `tool` message is read **off the wire**.
With the envelope branch removed:

```
not ok 1 - #331: an oversized Socrata envelope reaches the model as valid JSON carrying the truncation marker
  error: |-
    no row-truncation marker on the result fed to the model:
    …"},{"unique_key":"100337","created_date":"2026-01-01T00:00:00.000","complaint_type":"Noise - Residential","borough":"B
    [Truncated: result was 296028 characters]
```

That is #331 verbatim — cut mid-record, no `[Truncated: showing N of M rows]`. With the
fix, the body before the marker parses, `data.length` equals the marker's N, and
`total_rows` survives so the model still knows the size of the matching set upstream.
Three companion cases hold the other branches: a bare array still slices whole rows, a
non-JSON result keeps its character-truncation marker, and a result inside the bound is
fed back untouched.

**#343.** The `'Token budget exceeded'` arm is deleted. The output contract is now stated
exactly **once** — `FINAL_ANSWER_REQUEST`. The surviving note for a tool call the loop
never ran reports a fact and nothing else: *"This tool call was not run: the tool-call
limit for this query was reached."* With the old two-arm paraphrase restored:

```
not ok 9 - #343: only ONE message states the output contract, on the path that writes an abort note
  error: |-
    the abort note is stating the output contract a second time:
    Tool call limit reached. Please provide a summary based on the data already collected.
```

The paraphrase is the statement that goes unnoticed: it shares no wording with the
restatement, so counting words will not find it. What identifies it is that it tells the
model what to *write*, which is the restatement's job and no one else's — that is what the
test asserts. A second case pins P1's measurement that the deleted arm was unreachable:
on the token path the transcript carries the call's real result and no abort note at all.

### 6. #349 closes with all three properties, and `args` identity holds — PASS, shown red

Fixture: a tool call whose `function.arguments` is cut mid-string
(`{"type":"query","dataset_id":"sentinel-…`).

- **(a)** recorded with `failed: true`, `failureKind: 'unknown'`.
- **(b)** the model is told through `describeToolFailureForLlm` — the tool message is
  byte-equal to that function's output, and no request body anywhere contains
  `in JSON at position`, `Unterminated string`, `SyntaxError` or `Unexpected token`. The
  parse failure is raised as an app-authored Error rather than the `SyntaxError`, so
  endpoint-supplied bytes cannot steer `classifyStreamError` either. The **assistant**
  turn still carries the arguments the endpoint sent — that is the model's own turn, and
  rewriting it would make the transcript a fiction; the test asserts the sentinel appears
  there and nowhere else.
- **(c)** the run continues: a second, well-formed call executes and the query answers.

With the parse moved back outside the `try`:

```
not ok 5 - #349 (a): a malformed argument set is RECORDED as a failed tool call
not ok 6 - #349 (b): the model is told through describeToolFailureForLlm, with no parse-error text on the wire
not ok 7 - #349 (c): a malformed argument set does NOT end the run — the loop continues and answers
  error: 'Unterminated string in JSON at position 68 (line 1 column 69)'
  name: 'SyntaxError'
  stack: JSON.parse (<anonymous>) / runToolLoop (run-tool-loop.ts:541:50)
```

**`args` identity.** The probe asserts reference identity (`assert.equal`, i.e.
`Object.is`) *and* that a field injected inside the tool closure after the record was made
is visible on the record. With `args: { ...args }` on the record:

```
not ok 8 - the args on the record is the SAME object handed to executeToolCall
  error: the recorded args must BE the object the tool received
    + actual - expected
      { dataset_id: 'abcd-1234',
    -   portal: 'data.cityofnewyork.us',
        type: 'query' }
```

A deep-equality probe would have passed on that clone. This one does not.

**One extension beyond the criterion (`9edd841`), itemised.** `null`, a number or a bare
string is valid JSON: it clears the parse and then throws on the first property read one
step later, outside a guard that only watches the parse. It is the same failure and is now
handled identically. Shown red first — accepting any parse result gives
`Cannot read properties of null (reading 'type')` out of `deriveOperationType`:

```
not ok 8 - #349: JSON that parses but is not an argument set is the same failure
  error: "Cannot read properties of null (reading 'type')"
```

## The F-1 decision: `token_limit_exceeded`

**Reproduced, not changed.** The key is **present** on the answering-turn completion and
**absent** on the pass-through one, exactly as at `7357f27`. The two `onComplete` sites in
`queryWithMcpStreaming` stay distinct rather than being unified behind one adapter.

Reasoning: `tokenLimitExceeded` can only be true on a run that took the answering turn
(`abortedMidPlan` includes it), so on the pass-through path the flag would always be
`false` — and emitting `token_limit_exceeded: false` on every ordinary answer is a new
field on the SSE `complete` event for every query the app serves, read by a client that
today distinguishes "no flag" from "flag false". It is also newly pinned by a test
(`#345 P2: token_limit_exceeded is present … and ABSENT …`), which goes red under the
criterion-3 mutation, so the distinction cannot be flattened silently by a later phase.

## Gates

`npm ci` first. All five run at the pushed head.

**`npm test`**

```
1..1032
# tests 1065
# suites 10
# pass 1065
# fail 0
# cancelled 0
# skipped 0
# todo 0
# duration_ms 1811.249792
```

**`npm run build`** — fails in this sandboxed session, **re-measured rather than assumed**,
per the CLAUDE.md rule:

```
> Build error occurred
Caused by:
- creating new process
- binding to a port
exit 1
```

This session denies socket binding, so Turbopack's PostCSS worker pool dies. `--webpack`
skips it; CI's `build` job is the real gate.

**`npx next build --webpack`** — exit 0

```
▲ Next.js 16.3.0 (webpack)
- Environments: .env.local
✓ Running next.config.ts took 9ms

✓ Compiled successfully in 2.1s
  Running TypeScript ...
  Finished TypeScript in 1177ms ...
✓ Generating static pages using 9 workers (32/32) in 178ms
  Collecting build traces ...
Route (app)  … full table emitted, 32 routes …
```

**`npm run typecheck`** (run after the successful build) — exit 0, no output

```
> civic-ai-tools-website@0.1.0 typecheck
> tsc --noEmit
```

**`npm run lint`** — exit 0, the documented baseline, unchanged

```
scripts/eval-models.mjs
src/components/evidence/AttestationDialog.tsx
src/components/notebook/RenderingCellOutputs.tsx
  96:9  warning  Using `<img>` could result in slower LCP and higher bandwidth … @next/next/no-img-element
src/lib/bpmn/animation.ts

```

**`npm run check:compose-env`** — exit 0

```
```

## The D6 pre-authorized edit, with its own RED

`model-identity.test.ts:763-779` read `openrouter-streaming.ts` as text and asserted ≥2
`startSpan('llm_inference')` sites in it. Both sites moved into the core, so after
extraction that file has zero. Shown red before it was touched:

```
not ok 16 - #30 P6 F4: no inference span is started without gen_ai.system
  error: 'both inference sites should be found'
# tests 16 · # pass 15 · # fail 1
```

Replaced with the stronger tree-wide form P1 proposed and ORCH ruled for: it walks every
non-test `.ts`/`.tsx` under `src/`, collects `startSpan('llm_inference'` everywhere, and
asserts ≥2 sites each carrying `'gen_ai.system': getGenAiSystem()`. The now-false comment
("Both current sites are in this one file") is deleted. Green after:
`ok 16 … # pass 16 · # fail 0`.

Six, none of them fatal; three needed a decision.

1. **The allowlist in the contract is one entry short.** It names the core, the two
   unmigrated loops and the three single-call sites — six. `openrouter-streaming.ts` still
   calls `chat.completions.create` at what was `:313`: `queryWithoutMcpStreaming`, one
   turn, no tools. It is the A-side of the comparison and is not loop-class (the same
   status G0 item 4 gives `openrouter.ts:29`), so it is **allowlisted with that reason**
   rather than moved into the core. Seven entries, not six. Flagged rather than resolved
   silently because the count is the contract's.

2. **G1 axis 9 — "the pass-through chunker stays caller-side" — does not survive
   implementation.** The `synthesis` span *brackets* the chunker: it opens at the old
   `:597`, before the exit decision, and closes at `:746`, after the 20-chars-per-10 ms
   loop. Moving the chunker out of the core while leaving the span in it would move the
   span's recorded `endTimeUnixNano` earlier by roughly `content.length / 20 × 10 ms` — a
   change inside a signed trace, invisible in any test. The chunker therefore moved into
   the core as a **parameter** (`passThroughDelivery: { chunkChars, delayMs }`), which
   only the streaming caller passes; replay and compare omit it and get the content whole.
   The equivalence probe above shows the trace identical on every pass-through scenario.
   **P1 did not see this; proposed as the correction to axis 9.**

3. **One option beyond G1's signature: `logContext?: string`.** `responseModelAttributes`
   logs `[stream:${context}]` on a declared/reported model mismatch, and the streaming
   caller's context string is its panel. Without the option the operator log line would
   change wording. Defaults to `'model-loop'`.

4. **G1's `finalTurn` has no stated default; the core defaults to `'blocking'`.** The
   streaming caller passes `'stream'` explicitly. A non-streaming caller (P3, P4) needs no
   argument for the mode it wants, and `'blocking'` needs no `onDelta`.

5. **#331's fixture size, incidentally settled.** The 2,000-row envelope this phase builds
   is 296,028 characters — the anchor's and the issue's figure, not the 296,053 the G0
   record cited. Both were always plausible; they are different row shapes. No action.

6. **Contract line references: all verified exact at `7357f27`** — exit logic `:599-623`,
   transcript discipline `:398-406 :413 :579 :630-645`, `describeToolFailureForLlm`
   `:547-551`, failure record `:536-537`, `FINAL_ANSWER_REQUEST` `:231-237`,
   `announcesUnrunWork` `:212`, `truncateToolResult` `:273-293`, `MAX_TOKENS_PER_REQUEST`
   `:141`, spans `:371 :436 :567 :597`, chunker `:732-742`, `TraceContext` `:10-20`,
   `ProgressOpts` `:22-28`, harness `:657-728` with its driver `:730-763`,
   `model-identity.test.ts:763-779`. The two hard constraints were re-measured and both
   hold: `:421 → :424 → :447` is one object, and 0 of 74 test files import `@/`.

Two, both intended, both on the streaming path:

1. **#331** — an oversized paginated envelope now reaches the model as parseable JSON with
   a row marker instead of a fragment. Wire-visible; it is criterion 5.
2. **#349** — a tool call whose arguments will not parse no longer ends the query. Before,
   it escaped to `reportStreamFailure` and the reader got a sanitized failure for the
   whole run; now it is one failed tool call and the run continues. Wire-visible on the
   SSE `complete` event (an extra `tools_called[]` entry with `failed: true`) and on the
   absence of an `error` event.

`token_limit_exceeded` is **not** in this list — see the F-1 decision above.

## Flagged, not fixed

- **`mcp_tool_call` span attributes are built before `executeToolCall` runs**
  (`run-tool-loop.ts`, carried verbatim from `:436-443`), so a caller-injected `portal`
  reaches `tools_called[].args` but not `tool.portal_domain`. Pre-existing, unchanged,
  out of scope by the contract. Now that it lives in one place it is a one-line fix for
  whoever is next in the file — worth a phase of its own or a rider on P5.
- **`replay/route.ts` and `openrouter.ts` still carry their own loops**, with their own
  copies of #338/#344/#347/#349. Untouched, as the non-goals require. The registry test
  now names both with the phase that removes them, so the debt is visible in the suite
  rather than in an issue.
- **`src/lib/model-loop/test-harness.ts` is a non-test file that only tests import.** It
  is type-checked and linted like any source file and is not bundled by any route. P1
  recommended the extraction so P3 and P4 reuse the one instrument; naming it `.test.ts`
  would make the runner execute it as an empty suite.

## CI

All five checks pass at the pushed head **`9edd841`** — `build / test / lint / typecheck`,
`container image build`, `DCO`, `Vercel`, `Vercel Preview Comments` — and passed on
`059af3b` before it. `mergeable: MERGEABLE`.

## Riders honoured

One credentialed operation per Bash call — every commit and every push was its own
invocation, never bundled and never looped. `git commit -s`; the trailer matches the author exactly
(`Nathan Storey <npstorey@users.noreply.github.com>`). Pushed to `github-https`; the
global pre-push sensitivity hook ran and did not block, and nothing was bypassed. No
`.env*` file was read at any point; every key, model id and hostname in a fixture is an
obviously fake placeholder on a loopback address. No merge, no tag, no deploy.

### DIFF
--- a/src/lib/model-identity.test.ts
+++ b/src/lib/model-identity.test.ts
@@ -21,7 +21,9 @@
 import { test, beforeEach, afterEach } from 'node:test';
 import assert from 'node:assert/strict';
 import http from 'node:http';
-import { readFileSync } from 'node:fs';
+import { readFileSync, readdirSync } from 'node:fs';
+import { join, relative } from 'node:path';
+import { fileURLToPath } from 'node:url';
 import type { AddressInfo } from 'node:net';
 import canonicalize from 'canonicalize';
 import {
@@ -761,19 +763,47 @@ test('#30 P6 F4: every inference span carries gen_ai.system (measured on a trace
 });
 
 test('#30 P6 F4: no inference span is started without gen_ai.system', () => {
-  // The drift guard the trace test cannot give: a THIRD `llm_inference` span
-  // added later without the attribute would not show up in a fixture that
-  // never takes that branch. Both current sites are in this one file.
-  const source = readFileSync(
-    new URL('./openrouter-streaming.ts', import.meta.url),
-    'utf8',
+  // The drift guard the trace test cannot give: a `llm_inference` span added
+  // later without the attribute would not show up in a fixture that never
+  // takes that branch.
+  //
+  // This walks the whole tree rather than naming a file. It used to read
+  // `openrouter-streaming.ts` and say "both current sites are in this one
+  // file" — true when it was written, false the moment the tool-calling loop
+  // moved into `model-loop/run-tool-loop.ts` (#345), at which point a guard
+  // that named a file went red while the invariant it protects was untouched.
+  // A guard on a tree-wide property does not need editing when code moves.
+  const sites = inferenceSpanStarts();
+  assert.ok(
+    sites.length >= 2,
+    `expected at least two llm_inference span sites under src/, found ${sites.length}`,
   );
-  const starts = [...source.matchAll(/startSpan\('llm_inference'[\s\S]{0,240}?\}\)/g)];
-  assert.ok(starts.length >= 2, 'both inference sites should be found');
-  for (const match of starts) {
+  for (const site of sites) {
     assert.ok(
-      match[0].includes("'gen_ai.system': getGenAiSystem()"),
-      'an inference span without gen_ai.system breaks conformance with the declared version',
+      site.source.includes("'gen_ai.system': getGenAiSystem()"),
+      `an inference span without gen_ai.system breaks conformance with the declared version (${site.file})`,
     );
   }
 });
+
+/** Every `startSpan('llm_inference'…)` call site in non-test source under `src/`. */
+function inferenceSpanStarts(): { file: string; source: string }[] {
+  const srcRoot = fileURLToPath(new URL('..', import.meta.url));
+  return sourceFilesUnder(srcRoot).flatMap((file) => {
+    const source = readFileSync(file, 'utf8');
+    return [...source.matchAll(/startSpan\('llm_inference'[\s\S]{0,240}?\}\)/g)].map((match) => ({
+      file: relative(srcRoot, file),
+      source: match[0],
+    }));
+  });
+}
+
+function sourceFilesUnder(dir: string): string[] {
+  return readdirSync(dir, { withFileTypes: true }).flatMap((entry) => {
+    const full = join(dir, entry.name);
+    if (entry.isDirectory()) return sourceFilesUnder(full);
+    if (!/\.tsx?$/.test(entry.name)) return [];
+    if (/\.test\.tsx?$/.test(entry.name)) return [];
+    return [full];
+  });
+}
--- a/src/lib/model-loop/model-call-registry.test.ts
+++ b/src/lib/model-loop/model-call-registry.test.ts
@@ -1,0 +1,193 @@
+/**
+ * The registry: who is allowed to call the model directly (#345).
+ *
+ * WHY A TEST AND NOT A CONVENTION. Wave N6 fixed six defects and its cold read
+ * found three of them still alive elsewhere — each phase had fixed one copy of
+ * a duplicated defect while honouring its blast zone exactly. The largest
+ * family was the model-calling loop: three implementations, and every fix
+ * landed on one of them. Nothing in the repository could say "there are three
+ * of these" out loud, so nobody looked.
+ *
+ * This test says it out loud. Adding a `chat.completions.create` call to a
+ * file that is not on the list below fails the suite, and the list carries one
+ * line of justification per entry — including, for the two files this wave has
+ * not migrated yet, the phase that removes them.
+ *
+ * TWO ASSERTIONS, ONE INSTRUMENT. The first is the registry: who calls the
+ * model at all. The second is narrower and is the wave's own criterion: how
+ * many modules carry a TOOL-CALLING loop, which is the thing that was
+ * triplicated. A grep for `chat.completions.create` cannot answer the second
+ * (it counts single-turn calls too) and a grep for `tools` cannot answer it
+ * either (it hits every file that mentions the word), so the scanner below
+ * parses each call's argument block and asks whether that call passes tools.
+ *
+ * WHAT THE SCANNER DOES NOT DO. It reads source text; it does not build a call
+ * graph. A call assembled dynamically, or made through a helper that takes the
+ * arguments as a value, is invisible to it — the same limitation
+ * `ui-class-names.test.ts` and `design-tokens.test.ts` accept for their own
+ * scans. It is a drift guard against the pattern that actually recurs here
+ * (someone writes another loop), not a proof of exclusivity.
+ */
+
+import { test } from 'node:test';
+import assert from 'node:assert/strict';
+import { readdirSync, readFileSync } from 'node:fs';
+import { join, relative, sep } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const SRC_ROOT = fileURLToPath(new URL('../..', import.meta.url));
+
+/**
+ * Every file permitted to call `chat.completions.create`, with the reason it
+ * is on the list. An entry that names a phase is a debt with an owner; an
+ * entry that does not is a call this wave deliberately leaves where it is.
+ */
+const ALLOWED_MODEL_CALLERS: Record<string, string> = {
+  'src/lib/model-loop/run-tool-loop.ts':
+    'The shared tool-calling loop — the one implementation the wave consolidates onto.',
+  'src/lib/openrouter-streaming.ts':
+    'queryWithoutMcpStreaming: the no-tools A-side of the comparison. One turn, no loop, not loop-class.',
+  'src/app/api/evidence/[slug]/replay/route.ts':
+    'Its own loop, until P3 moves it onto the shared core (#338, #347).',
+  'src/lib/openrouter.ts':
+    'Its own loop, until P4 moves it onto the shared core (#344).',
+  'src/lib/evidence/adversarial-eval.ts':
+    'One rubric call, no loop. Out of this wave by ruling D4; the eval pair is #348.',
+  'src/app/api/evidence/[slug]/evaluate/route.ts':
+    'The second copy of that same rubric call, caller-keyed. Out of this wave by ruling D4 (#348).',
+  'src/app/api/evidence/generate-summary/route.ts':
+    'One summary call, no loop. Out of this wave.',
+};
+
+/**
+ * Modules carrying a tool-calling loop — a `chat.completions.create` that
+ * passes `tools`. Three at this wave's base; one when it closes.
+ */
+const ALLOWED_TOOL_LOOPS: Record<string, string> = {
+  'src/lib/model-loop/run-tool-loop.ts': 'The shared core.',
+  'src/app/api/evidence/[slug]/replay/route.ts': 'Removed by P3.',
+  'src/lib/openrouter.ts': 'Removed by P4.',
+};
+
+const MODEL_CALL = 'chat.completions.create';
+
+interface ModelCall {
+  file: string;
+  /** The text of the call's argument list, parens balanced. */
+  args: string;
+}
+
+test('#345: every model call site is on the registry, and every registry entry is a real call site', () => {
+  const calls = modelCallSites();
+  assert.ok(calls.length > 0, 'the scanner found no model calls at all — it has stopped measuring');
+
+  const callers = new Set(calls.map((c) => c.file));
+
+  for (const file of callers) {
+    assert.ok(
+      file in ALLOWED_MODEL_CALLERS,
+      `${file} calls ${MODEL_CALL} and is not on the registry in src/lib/model-loop/model-call-registry.test.ts. ` +
+        'Add it with one line saying why it is a separate call site — or, better, call runToolLoop.',
+    );
+  }
+
+  for (const file of Object.keys(ALLOWED_MODEL_CALLERS)) {
+    assert.ok(
+      callers.has(file),
+      `${file} is on the registry but no longer calls ${MODEL_CALL}. Remove the entry — a stale allowlist is a false statement about the tree.`,
+    );
+  }
+});
+
+test('#345: only the named modules carry a tool-calling loop', () => {
+  const loops = new Set(modelCallSites().filter(passesTools).map((c) => c.file));
+
+  for (const file of loops) {
+    assert.ok(
+      file in ALLOWED_TOOL_LOOPS,
+      `${file} calls ${MODEL_CALL} with tools — a second tool-calling loop. This is the shape wave #345 exists to end: call runToolLoop instead.`,
+    );
+  }
+  for (const file of Object.keys(ALLOWED_TOOL_LOOPS)) {
+    assert.ok(loops.has(file), `${file} is listed as carrying a tool loop but no longer does. Remove the entry.`);
+  }
+});
+
+/** True when this call passes `tools` — `tools,` shorthand or `tools:`. */
+function passesTools(call: ModelCall): boolean {
+  return /(^|[\s,{])tools\s*[,:]/.test(call.args);
+}
+
+function modelCallSites(): ModelCall[] {
+  return sourceFiles(SRC_ROOT).flatMap((path) => {
+    const source = readFileSync(path, 'utf8');
+    const file = `src/${relative(SRC_ROOT, path).split(sep).join('/')}`;
+    const calls: ModelCall[] = [];
+    let from = 0;
+    for (;;) {
+      const at = source.indexOf(MODEL_CALL, from);
+      if (at === -1) break;
+      from = at + MODEL_CALL.length;
+      if (onACommentLine(source, at)) continue;
+      calls.push({ file, args: argumentList(source, from) });
+    }
+    return calls;
+  });
+}
+
+/**
+ * True when the line holding `at` opens as a comment. Enough for this scan:
+ * the only non-call occurrence in the tree is a doc comment in
+ * `model-client.ts`, and a general comment stripper would have to reason about
+ * regex literals and URLs in string literals to do no worse.
+ */
+function onACommentLine(source: string, at: number): boolean {
+  const lineStart = source.lastIndexOf('\n', at) + 1;
+  const before = source.slice(lineStart, at).trimStart();
+  return before.startsWith('//') || before.startsWith('*') || before.startsWith('/*');
+}
+
+/**
+ * The text between the `(` at or after `from` and its matching `)`. String and
+ * template literals are skipped so a bracket inside one cannot unbalance the
+ * scan; nesting inside a template's `${}` is not tracked, which is fine
+ * because no call in this tree puts one in its argument list.
+ */
+function argumentList(source: string, from: number): string {
+  const open = source.indexOf('(', from);
+  if (open === -1) return '';
+  let depth = 0;
+  for (let i = open; i < source.length; i++) {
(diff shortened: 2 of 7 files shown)


npstorey:
Seat read of `9edd841`, three-dot against `main` `7357f27`: numstat +1827/−686 over 7 files = the PR API; merge-base is `7357f27`; consumer routes 0 lines; four required checks SUCCESS on this exact head; runner log for run 33111840394 reads `# tests 1065 · # pass 1065 · # fail 0` (1044 + 21). Test-name diff of my own (test/it/describe across every test file): 959 → 980, **nothing at base absent at head**. Read in the core, not the record: the #334 exit logic (`abortedMidPlan || answeredNothing || announcesUnrunWork`) at `:695-699`; one `FINAL_ANSWER_REQUEST` with the #343 arm replaced; zero clones/freezes, zero `PanelType`, zero `@/` imports; the #349 parse inside the try with `toolFailureKindOf` + `describeToolFailureForLlm`; the #331 envelope branch emitting `[Truncated: showing N of M rows]`; the registry test bidirectional with replay and openrouter.ts entered as "Removed by P3/P4". All six criteria hold.

**One thing to fix on top before the GO binds — commit provenance, not code.** Both commits carry the DCO trailer and no `Co-Authored-By`. Every N6 phase commit carries one naming the model that ran the phase (`2735644`, `fb1fbfb`, `525c559` → `Claude Opus 5 (1M context)`; `38660c4`, `9fc8c6b`, `a39718a` → `Claude Sonnet 5`), and that was the #325 G0 ruling: keep `Signed-off-by` and `Co-Authored-By`, drop `Claude-Session`, and the co-author string names the model that actually ran the phase. Merge-commit-only exists so per-commit provenance survives onto `main`; these two would survive without it. Amend both messages (message-only — tree identity should match on both sides, `git show -s --format=%T`), trailers in one contiguous block (`Signed-off-by` first, then `Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>`), one signing op per Bash call, force-push to the branch, post the new head. **GO re-binds to that head** once its four checks are green; a moved head voids this read. Put the trailer rule in P3's and P4's contracts alongside the DCO line.

npstorey:
**GO — head-bound to `8de9d2c`.** Verified by the seat: tree identity with the already-read `9edd841` (`git show -s --format=%T` equal on both, `git diff 9edd841 8de9d2c` empty), so the code read in comment 5445000795 carries unchanged; merge-base still `7357f27`; two commits, `ef0ece1` and `8de9d2c`, both sig G, each with `Signed-off-by` and `Co-Authored-By: Claude Opus 5 (1M context)` in one contiguous trailer block; required checks on this exact head — `build / test / lint / typecheck`, `container image build`, DCO, Vercel — all SUCCESS; `mergeStateStatus: CLEAN`. Closes #331, #343, #349. Owner's key: merge commit. A moved head voids this GO.

Issue #343, opened 2026-08-27
Unreachable abort branch, and two places restating the output contract


npstorey:
## Summary

An unreachable branch in the tool-calling loop, and a contract restatement it left behind.

## Measured at `1ddd61a`

`src/lib/openrouter-streaming.ts:635-637` selects an abort message with:

```ts
pendingToolCalls && !lastMessageAlreadyInTranscript
```

`tokenLimitExceeded` can only be set at `:558`, which is reached after `lastMessageAlreadyInTranscript = true` at `:413`, with nothing resetting it before the `break`. So when `tokenLimitExceeded` is true, the condition at `:634` is always false, and the `'Token budget exceeded. Please provide a summary based on the data already collected.'` arm can never be selected.

Verified by driving the loop: on the token-limit path the transcript is `[user] question → [assistant] +tool_calls → [tool] <real result> → [user] <final answer request>`, with no abort message.

## The surviving arm is live, and says something weaker

On the max-iterations path (measured: 22 model requests) the transcript carries **both**:

```
[tool]  Tool call limit reached. Please provide a summary based on the data already collected.
[user]  This is the final turn: no further tool calls will be made, …
```

The tool-message wording is the "just summarize" phrasing that the newer restatement was written to replace. They do not contradict each other today, but there are two places stating an output contract and editing one gives no reason to look at the other.

## Suggested direction

Remove the unreachable arm, and decide whether the abort message should defer to the single contract restatement rather than paraphrasing it.

npstorey:
Shipped in **PR #350**, merged as `a9681ab` (Wave N7 P2).

The unreachable `'Token budget exceeded'` arm is deleted, and the output
contract is now stated once. Measured before the change, on the two paths this
issue describes: the token-budget path produced **0** occurrences of that
wording (the arm could not be selected, exactly as filed), and the
max-iterations path produced 1 occurrence of `'Tool call limit reached'` at 22
model requests — the figure this issue reported.

The note for a tool call the loop never ran now reports that fact and says
nothing about how to answer, so there is one statement of the contract rather
than a paraphrase beside it. Pinned by two tests in
`src/lib/model-loop/run-tool-loop.test.ts` — one asserting a single contract
message on the path that writes an abort note, one asserting the token-limit
path writes no abort note at all, so the state the deleted arm required cannot
recur.

Closed: this defect was scoped to one file, and that file's loop no longer
exists — it is the shared core now.

Issue #342, opened 2026-08-27
The failed-call note can describe an operation that never happened


npstorey:
## Summary

The markdown note that replaces a rejected tool call can describe something that never happened, in a self-contradicting sentence, in published output.

## Measured at `1ddd61a`

`renderFetchToolCell` (`src/lib/notebook-author/tool-to-cell.ts`) dispatches on the tool **name and `args.type`** — `get_data` with `type: 'query'` is a data fetch; `get_data` with `type: 'catalog'` is a discovery search.

`describeAttempt` (`:335-354`), used only on the failed path at `:383`, dispatches on the **name alone**. A rejected catalog search therefore renders as:

> The original analysis tried to query the `unknown` dataset on `data.example.gov` to search for relevant datasets.

Three things are wrong in one sentence: no dataset was queried, there is no dataset called `unknown`, and the trailing reason clause contradicts the main clause.

## Why it matters

The note exists because an honest description of a refused call teaches the reader more than a cell that throws. That reasoning is right, and it is exactly why the note being wrong is worth fixing — a reader who trusts the note is misled about what the analysis attempted.

## Suggested direction

Give `describeAttempt` the same name-and-type dispatch the renderer uses, so a failed call is described by the operation it actually was.

npstorey:
Chartered in Wave N8 (#363, P5). Alive at 96c4f76, line-exact: describeAttempt at src/lib/notebook-author/tool-to-cell.ts:335-354 dispatches on the tool name alone, so every failed get_data call is described as a dataset query (:336, datasetId defaulting to unknown); it is reached only from renderFailedToolCell at :383.

npstorey:
Shipped in #370 (Wave N8 P5), merged as `f117665`, tag `rollback/pre-n8-p5-merged`.

`describeAttempt` now dispatches on the tool name **and** `args.type`, matching the dispatch `renderFetchToolCell` already used. A rejected catalog search is described as a search; it is no longer rendered as a query of a dataset called `unknown`, and the sentence no longer contradicts itself.

Shown red first: reverting `describeAttempt` to name-only dispatch fails both of the phase's #342 cases.

[...]

Issue #341, opened 2026-08-27
A notebook whose every fetch failed validates ok and still presents the original figures as its conclusion


npstorey:
## Summary

A notebook in which every fetch failed executes cleanly, validates `ok`, and still presents the original answer's figures as its conclusion — with nothing behind them.

Before fetch cells were guarded, the first failing cell aborted the run and the failure was loud. Guarding them was correct; this is the consequence nothing decided.

## What the artifact says versus what it contains

Constructed from the real renderers at `1ddd61a`. Cell layout: `markdown, code, code, code, markdown, markdown, markdown, code, code, markdown` — **no fetch cells, no discovery cell**, `dataFrameVariables: []`.

Meanwhile:

- The cover cell still reads *"This notebook contains a complete, reproducible analysis… the same numbers can be reproduced against live data by re-running cells top-to-bottom."*
- The synthesis cell is `try/except`-wrapped and falls back to `display(Markdown(<the chat-flow answer>))`, so **the original answer renders as the conclusion, carrying its original figures**, with no computation behind it.
- `validateExecutedNotebook` (`src/lib/notebook-author/validate.ts:140-145`) checks provenance and the execution extension only, so it reports `ok`.
- The "Not reproduced" note says *"The steps below rest only on the requests that did return data"* — true, and vacuous when none did.

## Why it matters

This is the same shape as the mid-plan-answer defect this wave was built around: **something that looks computed and is not**, presented with the visual treatment of a real result. It has arrived on the notebook surface as a side effect of a fix that was itself correct.

The Phase D comparison cell does print `dfN_n_rows: original=…, current=0` on **re-execution**, which is a genuine mitigation for the re-execution case. It does not help when every fetch fails during the publisher's own authoring run, which is when the artifact is created.

## Suggested direction

A notebook whose every fetch failed should not describe itself as a complete reproducible analysis, and its synthesis cell should not restate figures it did not compute. Either the cover text becomes conditional on at least one DataFrame existing, or validation gains a check that a notebook claiming reproducibility contains at least one successful fetch — or both.

npstorey:
Chartered in Wave N8 (#363, P5). Alive at 96c4f76, line-exact: src/lib/notebook-author/prompt.ts:81 asserts a complete, reproducible analysis unconditionally, and validateExecutedNotebook at validate.ts:140-145 merges provenance and execution-extension issues only - no fetch-success check exists.

npstorey:
[...]

PR #351, opened 2026-08-27
fix: replay runs on the shared loop core (#338, #347, #331)

### LINKED ISSUE #331: truncateToolResult feeds the model malformed JSON for any oversized Socrata envelope

### LINKED ISSUE #338: Replay route carries the mid-plan-answer defect and feeds a signed consistency attestation

### LINKED ISSUE #347: Replay's token-budget break duplicates an assistant turn and re-answers its tool calls

npstorey:
# P3 — replay on the shared loop core

**Branch** `sprint-345/p3-replay-on-core` · **head** `b8b3cedc4d61c7f558b60f3b630f82ef171b5c0b` · **base** `a9681ab` · rollback tag `rollback/pre-345-p3` (pre-existing). Ran on **Opus 5** (`claude-opus-5[1m]`).

Closes #338
Closes #347
Closes #331

## Diff and blast zone

```
 src/app/api/evidence/[slug]/replay/route.ts    | 164 ++---------
 src/lib/model-loop/model-call-registry.test.ts |   7 +-
 src/lib/model-loop/replay-loop.test.ts         | 392 +++++++++++++++++++++++++
 src/lib/model-loop/replay-loop.ts              | 122 ++++++++
 4 files changed, 542 insertions(+), 143 deletions(-)
```

`git diff --name-only a9681ab...HEAD` returns **exactly the four declared paths and nothing else**. `src/components/evidence/AttestationDialog.tsx` is **untouched** — see criterion 6 for why no edit was forced. Nothing out of zone was needed, so nothing is itemised as an out-of-zone need.

## What moved where

| Concern | Now lives in |
|---|---|
| the tool-calling loop, its exit condition, truncation, error-to-model path, transcript discipline | `src/lib/model-loop/run-tool-loop.ts` (P2's core, unchanged) |
| caps (20 iterations / 4000 `max_tokens` / 200k cumulative / 50k tool-result chars), the MCP tool set, the `portal` injection, the 45 s per-call race | `src/lib/model-loop/replay-loop.ts` — new, relative imports only |
| session check, caller key, MCP routing refusal, record + package reads, `canReadRecord`, hash-only refusal, `endpointModelForDeclared`, the `portal` derivation, `buildSystemPrompt(portal)`, `createModelClient`, the whole error-classification tail | `replay/route.ts`, unchanged |

The route's private loop and its private `truncateToolResult` are deleted. The route now holds no loop configuration at all; a test in `replay-loop.test.ts` asserts that (`replayLoopOptions(` present, no `callTool`, no `maxIterations` / `max_tokens` / `maxCumulativeTokens` / `truncateToolResult`).

`replayLoopOptions` exposes exactly one seam, `callTool`, one level **below** the loop. A test substitutes the tool transport and restates no loop configuration — change replay's caps and the tests change with it.

## Acceptance — six criteria, RED then GREEN

RED for criteria 2–6 was produced by a driver that **mechanically slices** the loop out of `a9681ab:src/app/api/evidence/[slug]/replay/route.ts` (byte ranges `:23-25`, `:27-42`, `:137-253`) via `git show` and runs it against the same scripted endpoint and the same fixtures as the new path. Nothing is hand-transcribed; the substitutions are the route wrapper (replaced by parameters), `NextResponse.json` (identity), and the `:254-276` classifier tail (dropped, caller-side by design). This is the same limitation G1 recorded: it proves the defect is in that code, not that the route file executes it. The registry test and the route drift guard close that gap from the other side.

### 1 — Replay drives the core; its loop and its truncation copy are gone

`replay/route.ts` contains no `chat.completions.create`, no tool-calling `while`, and no `truncateToolResult`. Both `model-call-registry.test.ts` entries removed. **Both directions of the bidirectional test were exercised:**

*RED direction A — entries left behind, call site gone:*

```
  error: 'src/app/api/evidence/[slug]/replay/route.ts is on the registry but no longer calls chat.completions.create. Remove the entry — a stale allowlist is a false statement about the tree.'
  error: 'src/app/api/evidence/[slug]/replay/route.ts is listed as carrying a tool loop but no longer does. Remove the entry.'
# tests 2 · # pass 0 · # fail 2
```

*RED direction B — entries removed, base route restored:*

```
  error: 'src/app/api/evidence/[slug]/replay/route.ts calls chat.completions.create and is not on the registry ... Add it with one line saying why it is a separate call site — or, better, call runToolLoop.'
  error: 'src/app/api/evidence/[slug]/replay/route.ts calls chat.completions.create with tools — a second tool-calling loop. This is the shape wave #345 exists to end: call runToolLoop instead.'
# tests 2 · # pass 0 · # fail 2
```

*GREEN at head:* `# tests 2 · # pass 2 · # fail 0`.

A third guard was added in `replay-loop.test.ts` — *"the replay route obtains its loop options from this factory"*. Shown red against the base route: `error: 'the route must build its options here, not inline'`.

### 2 — No announcement is published as replay's output (the wave's property-shaped criterion)

Fixture final turn, 137 characters, first person, a data verb:
`"I now have the counts by request type for both years. Next, I'll query the fraction of records that close within 14 and 30 days per type."`

```
RED  base a9681ab  output === the announcement: true
RED  base a9681ab  output: "I now have the counts by request type for both years. Next, I'll query the fraction of records that close within 14 and 30 days per type."
RED  base a9681ab  model requests made: 2

GREEN P3           output === the announcement: false
GREEN P3           output: "Across both years the portal recorded 4,812 requests of this type. The median time to close was 9 days, and 71% closed within 14 days (dataset abcd-1234)."
GREEN P3           model requests made: 3
```

Pinned by `#338: a final turn that announces a query is not returned as the replay output` and its companion `a genuine answer is published as written, with no extra model call` (2 requests, not 3 — the fix costs nothing on the common path).

### 3 — No raw error text reaches the model; failed calls are recorded as failed

A tool throwing `Error("ECONNREFUSED db.internal.example:5432")`, asserted **on the wire**:

```
RED  base a9681ab  tool message on the wire: "Error: ECONNREFUSED db.internal.example:5432"
RED  base a9681ab  wire contains "db.internal.example": true
RED  base a9681ab  record: {"name":"get_data","args":{"type":"query","dataset_id":"abcd-1234","portal":"data.cityofnewyork.us"}}
RED  base a9681ab  record has failed/failureKind: false / false

GREEN P3           tool message on the wire: "This data request returned no data. Do not estimate, guess, or fabricate any values to fill the gap. The live data source is temporarily unavailable. Suggest trying again shortly. In your answer, briefly tell the user in plain language that the live data could not be retrieved, and do not include any raw error text, status codes, server names, or system details."
GREEN P3           wire contains "db.internal.example": false
GREEN P3           record failed=true failureKind=unavailable
```

### 4 — #347 closes with a regression test

Token-budget path, final outgoing request:

```
RED  base a9681ab  final request: assistant turns = 2
RED  base a9681ab  final request: tool messages for call_1 = 2
RED  base a9681ab  their contents = ["{\"data\":[{\"request_type\":\"A\",\"days_to_close\":9}],\"total_rows\":1}","Limit reached. Please summarize based on data collected."]

GREEN P3           final request: assistant turns = 1
GREEN P3           final request: tool messages for call_1 = 1
GREEN P3           their contents = ["{\"data\":[{\"request_type\":\"A\",\"days_to_close\":9}],\"total_rows\":1}"]
```

P1's measurement reproduced exactly (2 turns, `call_1` answered twice) before the fix.

### 5 — #331's second instance is gone

A 2,000-row paginated envelope. Its length came out at **296,028 characters** — the issue's own figure, from an independently generated fixture.

```
envelope length: 296028

RED  base a9681ab  tool message length: 50042
RED  base a9681ab  carries a row marker: false
RED  base a9681ab  body is valid JSON: false — Unterminated string in JSON at position 50000 (line 1 column 50001)
RED  base a9681ab  last 120 chars: ":\"2026-01-01T00:00:00.000\",\"complaint_type\":\"Noise - Residential\",\"borough\":\"B\n[Truncated: result was 296028 characters]"

GREEN P3           tool message length: 49646
GREEN P3           carries a row marker: true
GREEN P3           body is valid JSON: true
GREEN P3           last 120 chars: "se - Residential\",\"borough\":\"BROOKLYN\",\"incident_zip\":\"11201\"}],\"total_rows\":2000}\n[Truncated: showing 335 of 2000 rows]"
```

50,042 characters out, and a parse failure at position 50000, matches #331's own measurement. The committed test asserts on the wire: marker present, body parses, `total_rows` survives, body within the bound.

### 6 — The attestation payload changes only additively

Keys computed with `canonicalizeToolCall` copied verbatim from `AttestationDialog.tsx:55-63`, over a four-call fixture that exercises injection, a portal already supplied, and a non-Socrata tool:

```
base a9681ab keys: ["get_data:catalog::data.cityofnewyork.us","get_data:query:erm2-nwe9:data.cityofnewyork.us","get_data:metrics:erm2-nwe9:data.sfgov.org","get_variables:::"]
P3           keys: ["get_data:catalog::data.cityofnewyork.us","get_data:query:erm2-nwe9:data.cityofnewyork.us","get_data:metrics:erm2-nwe9:data.sfgov.org","get_variables:::"]
byte-identical: true
```

The first two carry a `portal` **the endpoint never sent** — injected by `executeToolCall` into the object the core already recorded. That is the constraint the wave named, and it is asserted as literal expected strings in the committed test, not as "toolCalls is non-empty".

```
base a9681ab record[0] keys: ["name","args"]
P3           record[0] keys: ["name","args","operationType","reason","duration_ms","resultSummary"]
P3 adds only: ["operationType","reason","duration_ms","resultSummary"]   (plus failed/failureKind on a failed call)
P3 drops:     []

base payload top-level keys: ["toolCalls","output","tokenUsage","durationMs"]
P3   payload top-level keys: ["toolCalls","output","tokenUsage","durationMs"]
base tokenUsage: {"promptTokens":30,"completionTokens":15,"totalTokens":45}  durationMs typeof number
P3   tokenUsage: {"promptTokens":30,"completionTokens":15,"totalTokens":45}  durationMs typeof number
P3   output produced: true
```

**Why `AttestationDialog.tsx` needed no edit.** `ReplayResult` (`:33-38`) types a value that arrives from `await res.json()`, so extra runtime fields are not an excess-property error; `canonicalizeToolCall` reads four fields; and the *signed* payload (`:300-308`) carries only `toolCallCount`, `toolCallKeys`, `outputExcerpt`, `tokenUsage`, `durationMs` — none of which changes shape. The signed consistency attestation is byte-compatible except where `outputExcerpt` now holds an answer rather than an announcement, which is the fix.

## Gate checks — run after `npm ci`

```
### npm test
1..1042
# tests 1075
# suites 10
# pass 1075
# fail 0
# cancelled 0
# skipped 0
# todo 0
# duration_ms 2100.273208
```

The suite with `replay-loop.test.ts` removed reports **1065**, so this branch adds 10 tests and removes none. (1044 at `1ddd61a` and `7357f27`; P2 took it to 1065.)

```
### npm run build
> next build
Build error occurred
Caused by:
- creating new process
- binding to a port
```

**Re-measured, not assumed.** This session denies socket binding, so Turbopack's PostCSS worker pool dies — the documented environment failure, not a defect in this change. `--webpack` skips the worker pool:

```
### npx next build --webpack
▲ Next.js 16.3.0 (webpack)
- Environments: .env.local
✓ Running next.config.ts took 53ms

✓ Compiled successfully in 2.2s
  Running TypeScript ...
  Finished TypeScript in 1213ms ...
✓ Generating static pages using 9 workers (32/32) in 236ms
  Collecting build traces ...

Route (app) table emitted; both /api/evidence/[slug]/replay and /api/records/[slug]/replay present.
```

CI's `build` job is the real gate.

```
### npm run typecheck   (run after a successful build)
> tsc --noEmit
exit=0   (no output)
```

```
### npm run lint
> eslint

/Users/npstorey/code/civic-ai-tools-website/scripts/eval-models.mjs

/Users/npstorey/code/civic-ai-tools-website/src/components/evidence/AttestationDialog.tsx

/Users/npstorey/code/civic-ai-tools-website/src/components/notebook/RenderingCellOutputs.tsx
  96:9  warning  Using `<img>` could result in slower LCP and higher bandwidth. Consider using `<Image />` from `next/image` or a custom image loader to automatically optimize images. This may incur additional usage or cost from your provider. See: https://nextjs.org/docs/messages/no-img-element  @next/next/no-img-element

/Users/npstorey/code/civic-ai-tools-website/src/lib/bpmn/animation.ts

```

The documented baseline, unchanged: 4 warnings, 0 errors.

```
### npm run check:compose-env
```

No `.env*` file was read at any point.

**P3-1 — the client is constructed OUTSIDE the route's `try`, and stays there.** The contract said to construct it inside the existing `try` "so a `createModelClient` throw is classified exactly as today". At `a9681ab` it is at `:135` and the `try` opens at `:143`, so today a `createModelClient` throw is **not** classified — it escapes the handler entirely. Moving it in would have changed behaviour: a `ModelConfigurationError` classifies to `model_not_configured`, which `callerModelKeyFailure` deliberately returns `null` for, so it would have fallen through to a JSON 500 carrying the configuration message plus a `[replay] Error:` log line, where today the framework answers. Since the instruction's stated purpose was *no change*, the construction was left exactly where it was. Proposed correction, not made: classifying it is a small legibility win and belongs with whoever next revisits this route's error tail.

**P3-2 — G1's `replayLoopOptions` sketch predates two options P2 added.** Reconciled against the interface as built: the factory sets `finalTurn: 'blocking'` explicitly (the core's default, stated so the choice is visible to a reader) and `logContext: 'replay'`. It does not pass `passThroughDelivery` (no reader), `onDelta`/`onEvent` (nothing to report to), `trace` (replay writes no trace today) or `declaredModel` (only reachable through a trace).

**P3-3 — a rider that could not be honoured as written: `git commit -s` with the mandated block order.** Measured: `git commit -s` deduplicates the sign-off only when it is the **last** trailer line. With the mandated order (sign-off first, `Co-Authored-By` second) it appends a second `Signed-off-by`, which is what the first commit got. The commit was amended to the exact block the repo's merged commits carry — one sign-off first, one `Co-Authored-By` second, contiguous, author matching `npstorey@users.noreply.github.com`. Verified with `git log -1 --format='%(trailers:key=Signed-off-by,valueonly)|%(trailers:key=Co-Authored-By,valueonly)'`: one value each.

**Everything else in the contract verified exact at `a9681ab`:** `route.ts:23-25`, `:27-42` (the #331 gate at `:31`, the raw slice at `:40`), `:129`, `:132`, `:135`, `:137`, `:138`, `:149`, `:173`, `:174`, `:177-185`, `:195`, `:201`, `:203`, `:218`, `:221-229`, `:230`, `:238`, `:254-275`; `AttestationDialog.tsx:33-38`, `:55-63`, `:233`; the core's `:94-132` option members and its three defaults; `test-harness.ts`'s three exports; both "Removed by P3" registry entries; and #331's reopened state (the defective copy was still at `:27-42`).

Beyond the three defect fixes, four things a reader would otherwise discover rather than be told:

1. **The per-call timeout is now cleared.** The route's race left a 45-second `setTimeout` pending after every successful tool call (so do `compare-stream` and `query-notebook`). `replayLoopOptions` clears it in a `finally`. No payload effect; it is also what lets the test suite finish in 200 ms rather than 45 seconds per tool call.
2. **The timeout message gains the tool name** — `MCP tool "get_data" timed out after 45s` rather than `MCP tool timed out after 45s`, matching the other two callers. It classifies identically (`mcp_timeout`) and never reaches the model, which sees `describeToolFailureForLlm` copy.
3. **The unrun-tool-call note changes wording**, from `Limit reached. Please summarize based on data collected.` to the core's `This tool call was not run: the tool-call limit for this query was reached.` — #343's one-contract-stated-once, inherited from the core.
4. **A malformed tool-argument set is now one failed call rather than a 500** (#349, inherited from the core). On this route that matters more than elsewhere: it used to abort an N-run consistency test after N−1 replays had already been paid for.

`model-call-registry.test.ts` carries **one comment line beyond the two entries**: the header sentence "for the two files this wave has not migrated yet" became false with one file left, so it now reads "for the file this wave has not migrated yet". No other comment was touched — the `ALLOWED_TOOL_LOOPS` doc comment ("Three at this wave's base; one when it closes") is still true and was left alone.

## Flagged, not fixed

- **`openrouter.ts` still carries its own loop** with copies of #344, #349 and the pre-#334 exit condition. P4's, and now the only entry left in `ALLOWED_TOOL_LOOPS` besides the core.
- **`mcp_tool_call` span attributes are still built before `executeToolCall` runs**, so a caller-injected `portal` reaches `tools_called[].args` but not `tool.portal_domain`. Pre-existing, carried from P2's flag; replay passes no trace, so it is invisible on this path today. Still a rider for P5.
- **The replay route writes no trace at all.** It is the one signed-attestation-feeding path with no `llm_inference` or `mcp_tool_call` spans. Out of charter here; worth a decision rather than a silent continuation.
- **P3-1's classification gap** (above): a `createModelClient` misconfiguration on this route answers a framework 500 with no `[replay]` log line. Unchanged by this phase, deliberately.
- **`canonicalizeToolCall` now exists twice** — in `AttestationDialog.tsx:55-63` and copied verbatim into `replay-loop.test.ts`, with a comment saying why (the component is behind the `@/` alias `node --test` cannot resolve, and a shared helper would quietly agree with itself). The duplication is deliberate and load-bearing for criterion 6, but it is duplication and the next reader should know.
- **The RED driver lives only in the session scratchpad**, not in the repository. Its input is `git show a9681ab:…`, so it is reproducible from the rollback tag, but it is not runnable from a checkout of this branch.

## What was not touched

`src/lib/openrouter.ts` and `/api/compare` (P4); `AttestationDialog.tsx`; the eval pair (#348); the notebook paths (#341, #342); version bumps, CHANGELOG headings, publishing. No merge, no tag, no deploy.

### DIFF
--- a/src/app/api/evidence/[slug]/replay/route.ts
+++ b/src/app/api/evidence/[slug]/replay/route.ts
@@ -6,7 +6,8 @@ import {
   resolveCallerModelKey,
 } from '@/lib/caller-model-key';
 import { endpointModelForDeclared } from '@/lib/model-resolver';
-import type { ChatCompletionMessageParam } from 'openai/resources/chat/completions';
+import { runToolLoop } from '@/lib/model-loop/run-tool-loop';
+import { replayLoopOptions } from '@/lib/model-loop/replay-loop';
 import { getServerSession } from 'next-auth';
 import { authOptions } from '@/lib/auth';
 import { db } from '@/lib/db';
@@ -14,32 +15,9 @@ import { evidenceRecords } from '@/lib/db/schema';
 import { eq } from 'drizzle-orm';
 import { getPackage } from '@/lib/storage';
 import { canReadRecord } from '@/lib/evidence/sealed-access';
-import { mcpTools } from '@/lib/mcp/tools';
-import { callMcpTool } from '@/lib/mcp/client';
 import { getMissingMcpRoutingError } from '@/lib/mcp/registry';
 import { buildSystemPrompt } from '@/lib/mcp/socrata-skill';
 import type { EvidencePackage } from '@/lib/evidence/packager';
-
-const MAX_TOOL_RESULT_CHARS = 50_000;
-const MAX_ITERATIONS = 20;
-const MCP_TOOL_TIMEOUT_MS = 45_000;
-
-function truncateToolResult(result: string): string {
-  if (result.length <= MAX_TOOL_RESULT_CHARS) return result;
-  try {
-    const parsed = JSON.parse(result);
-    if (Array.isArray(parsed) && parsed.length > 0) {
-      const sampleRow = JSON.stringify(parsed[0]);
-      const rowSize = sampleRow.length + 2;
-      const maxRows = Math.max(5, Math.floor(MAX_TOOL_RESULT_CHARS / rowSize));
-      const truncated = parsed.slice(0, maxRows);
-      return JSON.stringify(truncated) +
-        `\n[Truncated: showing ${truncated.length} of ${parsed.length} rows]`;
-    }
-  } catch { /* fall through */ }
-  return result.slice(0, MAX_TOOL_RESULT_CHARS) +
-    `\n[Truncated: result was ${result.length} characters]`;
-}
 
 /**
  * POST /api/evidence/[slug]/replay
@@ -134,122 +112,32 @@ export async function POST(
   // Create a model client with the user's API key
   const openrouter = createModelClient({ apiKey: callerKey.apiKey });
 
-  const startTime = Date.now();
-  const toolsCalled: { name: string; args: Record<string, unknown> }[] = [];
-  let cumulativeTokens = 0;
-  let cumulativePromptTokens = 0;
-  let cumulativeCompletionTokens = 0;
-
   try {
-    const messages: ChatCompletionMessageParam[] = [
-      { role: 'system', content: systemPrompt },
-      { role: 'user', content: prompt },
-    ];
+    // The shared tool-calling loop (#345). Replay carried its own copy until
+    // P3: the same loop, with its own exit condition, its own truncation and
+    // its own error-to-model path — and each of those was a filed defect
+    // (#338, #347, #331) that had already been fixed on the other copy.
+    // Everything the loop is GIVEN lives in `replayLoopOptions`; nothing about
+    // how it runs is decided here.
+    const result = await runToolLoop(replayLoopOptions({
+      client: openrouter,
+      endpointModel: model,
+      prompt,
+      systemPrompt,
+      portal,
+    }));
 
-    let response = await openrouter.chat.completions.create({
-      model,
-      messages,
-      tools: mcpTools,
-      tool_choice: 'auto',
-      max_tokens: 4000,
-    });
-
-    cumulativeTokens += response.usage?.total_tokens || 0;
-    cumulativePromptTokens += response.usage?.prompt_tokens || 0;
-    cumulativeCompletionTokens += response.usage?.completion_tokens || 0;
-
-    let iterations = 0;
-
-    // Tool-calling loop
-    while (response.choices[0]?.message?.tool_calls && iterations < MAX_ITERATIONS) {
-      const assistantMessage = response.choices[0].message;
-      messages.push(assistantMessage);
-
-      for (const toolCall of assistantMessage.tool_calls!) {
-        if (toolCall.type !== 'function') continue;
-
-        const args = JSON.parse(toolCall.function.arguments);
-        // Socrata's get_data expects a portal; Data Commons tools don't — only inject for Socrata.
-        if (toolCall.function.name === 'get_data' && !args.portal) args.portal = portal;
-        toolsCalled.push({ name: toolCall.function.name, args });
-
-        try {
-          const result = await Promise.race([
-            callMcpTool(toolCall.function.name, args),
-            new Promise<never>((_, reject) =>
-              setTimeout(
-                () => reject(new Error(`MCP tool timed out after ${MCP_TOOL_TIMEOUT_MS / 1000}s`)),
-                MCP_TOOL_TIMEOUT_MS,
-              ),
-            ),
-          ]);
-          messages.push({
-            role: 'tool',
-            tool_call_id: toolCall.id,
-            content: truncateToolResult(result),
-          });
-        } catch (error) {
-          messages.push({
-            role: 'tool',
-            tool_call_id: toolCall.id,
-            content: `Error: ${error instanceof Error ? error.message : 'Unknown error'}`,
-          });
-        }
-      }
-
-      // Token budget check
-      if (cumulativeTokens >= 200_000) break;
-
-      response = await openrouter.chat.completions.create({
-        model,
-        messages,
-        tools: mcpTools,
-        tool_choice: 'auto',
-        max_tokens: 4000,
-      });
-
-      cumulativeTokens += response.usage?.total_tokens || 0;
-      cumulativePromptTokens += response.usage?.prompt_tokens || 0;
-      cumulativeCompletionTokens += response.usage?.completion_tokens || 0;
-      iterations++;
-    }
-
-    // Extract final output
-    let output = response.choices[0]?.message?.content || '';
-
-    // If still pending tool calls, force a final response
-    if (!output && response.choices[0]?.message?.tool_calls) {
-      messages.push(response.choices[0].message);
-      for (const tc of response.choices[0].message.tool_calls) {
-        messages.push({
-          role: 'tool',
-          tool_call_id: tc.id,
-          content: 'Limit reached. Please summarize based on data collected.',
-        });
-      }
-      const finalResponse = await openrouter.chat.completions.create({
-        model,
-        messages: [
-          ...messages,
-          { role: 'user', content: 'Based on the data collected, please provide a comprehensive answer.' },
-        ],
-        max_tokens: 4000,
-      });
-      output = finalResponse.choices[0]?.message?.content || '';
-      cumulativeTokens += finalResponse.usage?.total_tokens || 0;
-      cumulativePromptTokens += finalResponse.usage?.prompt_tokens || 0;
-      cumulativeCompletionTokens += finalResponse.usage?.completion_tokens || 0;
-    }
-
+    // The consistency-test payload, unchanged in shape. `toolCalls` entries
+    // now additionally carry `operationType`, `reason`, `resultSummary`,
+    // `duration_ms` and — for a call that failed — `failed`/`failureKind`
+    // (#338). The identity fields the client keys a run on
+    // (`name`, `args.type`, `args.dataset_id`, `args.portal`, the last of
+    // which this route injects) are byte-identical to what it read before.
     return NextResponse.json({
-      toolCalls: toolsCalled,
-      output,
-      tokenUsage: {
-        promptTokens: cumulativePromptTokens,
-        completionTokens: cumulativeCompletionTokens,
-        totalTokens: cumulativeTokens,
-      },
-      durationMs: Date.now() - startTime,
+      toolCalls: result.toolCalls,
+      output: result.content,
+      tokenUsage: result.usage,
+      durationMs: result.durationMs,
     });
   } catch (error) {
     console.error('[replay] Error:', error);
--- a/src/lib/model-loop/model-call-registry.test.ts
+++ b/src/lib/model-loop/model-call-registry.test.ts
@@ -10,8 +10,8 @@
  *
  * This test says it out loud. Adding a `chat.completions.create` call to a
  * file that is not on the list below fails the suite, and the list carries one
- * line of justification per entry — including, for the two files this wave has
- * not migrated yet, the phase that removes them.
+ * line of justification per entry — including, for the file this wave has not
+ * migrated yet, the phase that removes it.
  *
  * TWO ASSERTIONS, ONE INSTRUMENT. The first is the registry: who calls the
  * model at all. The second is narrower and is the wave's own criterion: how
@@ -47,8 +47,6 @@ const ALLOWED_MODEL_CALLERS: Record<string, string> = {
     'The shared tool-calling loop — the one implementation the wave consolidates onto.',
   'src/lib/openrouter-streaming.ts':
     'queryWithoutMcpStreaming: the no-tools A-side of the comparison. One turn, no loop, not loop-class.',
-  'src/app/api/evidence/[slug]/replay/route.ts':
-    'Its own loop, until P3 moves it onto the shared core (#338, #347).',
   'src/lib/openrouter.ts':
     'Its own loop, until P4 moves it onto the shared core (#344).',
   'src/lib/evidence/adversarial-eval.ts':
@@ -65,7 +63,6 @@ const ALLOWED_MODEL_CALLERS: Record<string, string> = {
  */
 const ALLOWED_TOOL_LOOPS: Record<string, string> = {
   'src/lib/model-loop/run-tool-loop.ts': 'The shared core.',
-  'src/app/api/evidence/[slug]/replay/route.ts': 'Removed by P3.',
   'src/lib/openrouter.ts': 'Removed by P4.',
 };
 
--- a/src/lib/model-loop/replay-loop.test.ts
+++ b/src/lib/model-loop/replay-loop.test.ts
@@ -1,0 +1,392 @@
+// Acceptance tests for the replay path on the shared loop core (#345 P3).
+//
+// WHAT IS UNDER TEST, and why it is this and not the route. `replay/route.ts`
+// is a Next route handler: `node --test` cannot invoke one. So the loop
+// configuration was moved OUT of the route into `replayLoopOptions`, and these
+// tests drive the real `runToolLoop` with the real options factory against a
+// local scripted model server. The only thing substituted is the tool
+// transport — one level BELOW the loop — so no case here restates a cap, a
+// budget, a tool set or the portal injection. Change replay's configuration
+// and these tests change with it; that is the point of the seam.
+//
+// A source-drift guard at the bottom closes the remaining gap: it asserts the
+// route actually obtains its options from this factory and supplies no
+// transport of its own. Between that and `model-call-registry.test.ts` (which
+// fails if the route calls the model at all), "the route runs this" is
+// measured rather than assumed.
+//
+// Every claim about what the model was SENT is asserted against the request
+// bodies the mock server received, because that is where the claims live: a
+// truncation that hands the model malformed JSON (#331), a raw error string in
+// a `tool` message (#338), and a duplicated assistant turn (#347) are all
+// invisible from the return value.
+//
+// No live endpoint, no credential, no MCP server. Every key value is an
+// obviously fake fixture and the address is loopback.
+//
+// Run with: npm test
+//   (or: node --test --experimental-strip-types src/lib/model-loop/replay-loop.test.ts)
+
+import { test, beforeEach, afterEach } from 'node:test';
+import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
+import { fileURLToPath } from 'node:url';
+import { runToolLoop, type ToolCallRecord } from './run-tool-loop.ts';
+import {
+  replayLoopOptions,
+  REPLAY_MAX_ITERATIONS,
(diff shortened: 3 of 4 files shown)


npstorey:
**GO — head-bound to `b8b3ced`.** Seat read, three-dot against `main` `a9681ab` (merge-base `a9681ab`): +542/−143 over 4 files = the PR API; `AttestationDialog.tsx` 0 lines; the route at head has 0 `chat.completions.create`, 0 `truncateToolResult`, 0 tool-loop `while`, and constructs the client at `:113` outside the `try` at `:115` — position preserved, as the corrected contract asked; `ALLOWED_TOOL_LOOPS` is now the core plus `openrouter.ts` ("Removed by P4"). Read in `replay-loop.ts`: replay's own caps as parameters, `portal` injected by mutating the same `args` object (identity kept), the 45 s race cleared in a `finally`. Tests read: #338 ×3, #347's well-formed-transcript case, #331's envelope case, the literal-key attestation-identity case with `canonicalizeToolCall` copied rather than imported, the drift guard that the route obtains its options from the factory. Runner log for run 33119705068: `# tests 1075 · # pass 1075 · # fail 0` (1065 + 10); my own test-name diff 980 → 990, nothing absent at head. One commit, sig G, `Signed-off-by` then `Co-Authored-By: Claude Opus 5 (1M context)` contiguous. Four required checks SUCCESS on this exact head; `CLEAN`. #352's claim verified at `main` (0 `clearTimeout` in either streaming caller) and its phrasing clean. Closes #338, #347, #331. Owner's key: merge commit. A moved head voids this GO.

Issue #359, opened 2026-08-28
A tool call's span and its recorded args disagree about the portal, inside the signed trace


npstorey:
## Summary

A tool call's span and its recorded arguments give two different accounts of the
same call, under one signature. The span attributes are built **before**
`executeToolCall` runs, so anything the caller injects inside that closure —
notably `portal` — reaches the tool-call record but never the span.

## Measured at `45aa6c0`

Driving `queryWithMcpStreaming` with a trace and the portal-injection closure
`src/app/api/compare-stream/route.ts:216` uses:

```
record.args        -> {"type":"query","dataset_id":"abcd-1234","portal":"data.traced.example"}
span tool.arguments-> {"type":"query","dataset_id":"abcd-1234"}
span tool.portal_domain -> undefined
```

The attributes are assembled in `src/lib/model-loop/run-tool-loop.ts` at the
`mcp_tool_call` span start, which precedes the `await executeToolCall(...)`.
`executeToolCall` is the documented seam for exactly these caller-side concerns,
so anything correct about using it produces this divergence.

## Why it matters

The trace goes **inside the signed package**. A reader checking a published
record can see `tools_called[].args` naming the portal that was queried, and the
span for the same call naming no portal at all. Neither is wrong on its face;
together they are inconsistent, and the reader has no way to tell which account
the run actually used.

It is not a wrong answer and no published figure changes. It is two accounts of
one event under one signature, which is the property a record package exists to
make checkable.

## Suggested direction

Set the source-identifying attributes after `executeToolCall` returns — the span
is ended after the call regardless, so the attributes can be attached at
`endSpan` alongside the result attributes rather than at `startSpan`. Keep
whatever must be known up front (the tool name) where it is.

## Provenance

Carried through Wave N7 as a known pre-existing item from P2 and P3, deliberately
not fixed inside those phases' blast zones, and confirmed with the measurement
above by the wave's cold read (#345 P5).

npstorey:
[...]

Issue #358, opened 2026-08-28
CLAUDE.md's loop rule and MAX_TOOL_RESULT_CHARS's documented scope no longer match the tree


npstorey:
Two places that told a reader where the model-calling loop lives, or how to tune
it, stopped being true when Wave N7 (#345) consolidated the loop.

## 1. `CLAUDE.md`'s standing rule now misdirects — and its grep has the blind spot

`CLAUDE.md:126-131` still reads:

> Besides `src/lib/openrouter-streaming.ts`, the model is called from
> `openrouter.ts`, `evidence/[slug]/replay`, … Before fixing anything in the
> answer path, `grep -rl "chat.completions.create" src` …

Measured at `45aa6c0`:

- `src/app/api/evidence/[slug]/replay/route.ts` contains **zero**
  `chat.completions.create`.
- `src/lib/openrouter.ts` has one, no-tools, in `queryWithoutMcp`.
- The rule names neither `src/lib/model-loop/run-tool-loop.ts` — the one module
  that now carries a tool loop — nor
  `src/lib/model-loop/model-call-registry.test.ts`, the test that supersedes the
  grep.

The rule was added because a file-scoped blast zone could not see a class-scoped
defect. It now sends a reader to files that no longer hold the code, and the grep
it prescribes is scoped to `src` — the exact scope that misses the fourth loop in
`scripts/eval-models.mjs` (#356).

## 2. A documented tuning knob reaches one of three callers

`docs/deploy.md:583-584` lists `MAX_TOOL_RESULT_CHARS` among the instance tuning
knobs with no scope qualifier.

Only `src/lib/openrouter-streaming.ts:101` reads it. `compare-loop.ts:56` and
`replay-loop.ts:42` hardcode `50_000`.

Pre-existing in kind — `/api/compare` had no bound at all before the wave — but
the consolidation introduced two new named constants that shadow a documented
variable, at exactly the moment a reader would expect one implementation to mean
one setting. Setting `MAX_TOOL_RESULT_CHARS` on an instance today silently does
nothing to two of the three callers.

Related: the bound itself does not hold on heterogeneous rows (#355), so this
should be settled alongside that fix rather than before it.

## Suggested direction

Rewrite the `CLAUDE.md` rule to name the shared core and the registry test as the
guard, and either widen the grep it prescribes past `src/` or drop the grep in
favour of running the registry test. Then decide whether
[...]

Issue #356, opened 2026-08-28
A fourth tool loop lives in scripts/eval-models.mjs, where neither the census nor the registry test can see it


npstorey:
## Summary

`scripts/eval-models.mjs` runs a fourth copy of the model tool-calling loop,
carrying every defect Wave N7 (#345) just removed from the other three. Neither
the wave's census nor the registry test that now guards the class can see it,
because both are scoped to `src/`.

## Measured at `45aa6c0`

Three `chat.completions.create` calls — `:432`, `:505`, `:571` — the first two
passing `tools`; `maxIterations = 10` at `:453`.

| Defect class | Where |
|---|---|
| Loop exit keyed on content being absent (the pre-#334 condition, #319) | `:557` |
| Raw error text fed to the model | `:484` — `` content: `Error: ${error.message}` `` |
| `{name, args}`-only tool records (#321's shape) | `:471` |
| No truncation of tool results at all | — |
| The superseded final-answer wording | `:577` |
| Unguarded argument parse (#349) — and here it is **outside any try/catch** | `:466` |

It also injects the portal at `:469` **without** the `name === 'get_data'` guard
every in-`src` copy carries, so a Data Commons tool call receives a Socrata
portal.

## Why nothing caught it

`src/lib/model-loop/model-call-registry.test.ts` roots its scan at `SRC_ROOT`
(`:39`) and accepts only `.ts` / `.tsx` (`:189`). A `.mjs` file outside `src/`
fails both filters.

That makes the test's own header claim at `:11-13` — *"Adding a
`chat.completions.create` call to a file that is not on the list below fails the
suite"* — false as written, with a live counter-example in the tree.

The wave's census used `git grep … -- 'src/**'`, so it inherited the same blind
spot. This is the wave's own lesson turned on the wave: a check scoped to a
directory cannot see a defect scoped to a behaviour.

## Severity

Below the in-app copies. It writes `scripts/eval-results-*.json` for manual
scoring; nothing it produces is signed or user-facing.

It matters because of what it is *for*: choosing which models this platform
offers. A harness that can record a statement of intent as a model's answer, and
that dies on the first malformed argument set, is scoring models on the exact
defects the application just spent a wave removing.

## Suggested direction

Two parts, and the second matters more than the first:

1. Decide what this script should be — ported onto `runToolLoop`, or explicitly
   declared a throwaway harness that is not held to the app's guarantees. If the
   latter, say so in the file, because it does not currently read that way.
[...]

Issue #354, opened 2026-08-28
Two /api/compare docs describe behaviour the code does not have, including a log line that cannot appear


npstorey:
Two documentation claims about the `/api/compare` path do not match the code.
Found during Wave N7 P4 (PR #353) while moving that route onto the shared loop
core; both are **pre-existing** and neither was caused by that change.

## 1. The operator is told to watch for a log line that cannot appear

`docs/instance-setup.md:368-373`, immediately after the §3 smoke test that
`curl`s `POST /api/compare`:

> Watch the server log as well as the response. A line reading
> `endpoint reported a different model than this instance declares` means your
> `model` field and what the endpoint answered with disagree.

That line is emitted by `responseModelAttributes`, and every call to it sits
inside a trace guard:

```ts
if (llmSpanId) {
  trace!.builder.endSpan(llmSpanId, {
    …
    ...responseModelAttributes(declaredModel, response.model, logContext),
  });
}
```

`llmSpanId` is assigned only when a `trace` is supplied. **`/api/compare` supplies
none** — it did not before this wave and does not after it. So on the one path
the setup guide points at, the warning has never been emitted, and an operator
following the instruction learns nothing from its absence.

This is the shape of a check that cannot fail: absence of the warning and absence
of a mismatch are indistinguishable, and the doc presents the first as evidence
of the second.

Either direction is a fix, and they are not equivalent:

- **Point the instruction at a path that actually traces** (the streaming query
  surfaces), and say so; or
- **Give `/api/compare` a trace** — in which case `declaredModel` must be passed
  to `compareLoopOptions`. It is omitted today precisely because the route writes
  no trace, so the core's default (the wire string) is inert. Add a trace without
  it and the span records the deployment alias as the declared identity, which is
  the defect website#30 P3 fixed elsewhere.

## 2. A stale design sketch of a module that no longer looks like that

`docs/project-plan.md:466-500` carries a January sketch of `lib/openrouter.ts`
showing a `queryWithMcp` that no longer exists — its loop moved onto the shared
core in PR #353, leaving the file with `queryWithoutMcp` only. The sketch also
hardcodes an OpenRouter base URL and `OPENROUTER_API_KEY`, an endpoint assumption
`src/lib/model-client.ts` retired: this instance calls whatever
chat-completions endpoint it is configured for, and the key is read under a
vendor-neutral name.

PR #353 updated the `POST /api/compare` **response** block in the same file
[...]

PR #360, opened 2026-08-28
fix: the bound bounds, and two instruments stop overstating (#355, #357, #356)

### LINKED ISSUE #355: truncateToolResult's character bound does not bound: row count is sized from rows[0] alone
## Summary

`truncateToolResult`'s character bound does not bound anything when a result's
rows differ in size. The number of rows to keep is computed from **`rows[0]`
alone**, and the assembled string is never re-checked against the budget.

Introduced by Wave N7 (#345) while closing #331. Before that wave a Socrata
envelope fell through to raw character truncation, which respected the bound and
produced invalid JSON — that was #331. The envelope branch added to fix it
produces valid JSON and no longer respects the bound. The trade was not visible
at the time and is not covered by any test.

## The code

`src/lib/model-loop/run-tool-loop.ts:393-398`

```ts
function keepRowsWithinBudget(rows: unknown[], maxChars: number): unknown[] {
  const sampleRow = JSON.stringify(rows[0]);
  const rowSize = (sampleRow ? sampleRow.length : 0) + 2; // comma + newline
  const maxRows = Math.max(5, Math.floor(maxChars / Math.max(rowSize, 1)));
  return rows.slice(0, maxRows);
}
```

Both structured branches call it — the bare array at `:372` and the
`{data, total_rows}` envelope at `:380`.

The file's own stated contract at `:361-363` is *"a body the model can read,
followed by one `[Truncated: ...]` line saying what was dropped. The character
bound is on the body, not on the marker."* The body is not bounded.

## Measured, 50,000-character budget

Functions transcribed verbatim from `:365-398`, driven with 3,000-row envelopes:

| Rows | Body fed back to the model | Over budget by | Marker |
|---|---|---|---|
| homogeneous | 50,055 | 55 | showing 144 of 3000 |
| first row half-size (mild skew) | **206,740** | 156,740 | showing 595 of 3000 |
| sparse first row, rest full | **871,116** | 821,116 | showing 2500 of 3000 |
| 3 rows × 100 KB | **300,113** | 250,113 | **showing 3 of 3 rows** |

Every body is valid JSON, so #331 stays fixed. The bound is what broke.

The last row is a second defect in the same output: **nothing was dropped**, and
the marker says rows were. A model told 3 of 3 rows were truncated may caveat a
complete answer, or re-query for data it already has.

## Why the shapes are ordinary, not contrived

Socrata's SODA JSON **omits null columns per row**, so a record with few
populated fields sitting next to fully-populated ones is a normal response, not
an edge case. `src/lib/mcp/client.ts` applies no size cap of its own, so the raw
envelope arrives whole. The mild-skew row needs no assumption about
null-omission at all and still breaches the bound four times over.

## Consequence

`MAX_TOOL_RESULT_CHARS` exists to bound input token growth. On these shapes it
does not, so the next model request carries the oversized body: a
context-length rejection that loses the query, or on a large-context endpoint a
very large billed call.

Worst placement is `/api/compare` — anonymous, rate-limited, and
`compareLoopOptions` sets no `maxCumulativeTokens`, so there is no second bound
behind the first.

## Why the suite does not catch it

`replay-loop.test.ts:272` and `compare-loop.test.ts:332` both assert
`body.length <= maxChars` — real assertions, but every fixture uses homogeneous
rows, the one shape where the computation happens to be right. The assertions
are true by construction of the fixture.

## Suggested direction

Size by the assembled result rather than by a sample: keep adding rows while the
serialized output stays within budget, or compute against an average and then
re-check and trim. Emit the `[Truncated: ...]` marker only when rows were
actually dropped. A regression test needs **heterogeneous** rows — a homogeneous
fixture cannot fail.

## Provenance

Found by Wave N7's cold read (#345 P5) at merged head `45aa6c0`, and reproduced
independently against the same head before filing.

### LINKED ISSUE #357: The replay timeout test passes with the timeout deleted
## Summary

`src/lib/model-loop/replay-loop.test.ts:357` is named *"a tool call that never
settles fails on replay's own timeout, and the run survives"* and comments that
it *"Proves the race moved with the configuration rather than being lost with the
loop."*

It proves neither. Deleting the entire timeout race leaves the test green.

## Measured at `45aa6c0`

Mutation: replace the `Promise.race` + timer block in
`src/lib/model-loop/replay-loop.ts:104-119` with `return callTool(name, args);`.

Result: **all 10 replay tests stayed green**, this one included.

The test drives its fake transport to throw an error whose *text* reads
`timed out after 45s`. What it actually exercises is `classifyStreamError`'s
text-to-kind mapping — not the race, not the timer, not the 45-second bound.

## Why it matters

The race is the only thing standing between a hung MCP call and a replay run
that never returns — and a replay run is one of N feeding a signed consistency
attestation. Its removal, deliberate or accidental, would be caught by nothing
in the suite, while a test named for it reports green.

This is the shape the wave that introduced it was chartered to eliminate:
absence and success look identical. It is worth fixing precisely because someone
reading the suite would reasonably believe the behaviour is covered.

## Suggested direction

Drive a transport that genuinely never settles, with an injectable timeout so the
test does not wait 45 seconds — the bound is already a module constant and could
take an override the same way `callTool` does. Then verify the mutation above
turns the test red before trusting it.

## Provenance

Found by Wave N7's cold read (#345 P5) at merged head `45aa6c0`, by mutation
rather than by reading. The mutation was reverted; the tree is unchanged.

npstorey:
Wave N7 (#345) Phase 6 — the fix phase the G5 cold read opened. Three things: a
defect the wave **introduced** while closing #331, a new test that cannot fail,
and an instrument whose claim is wider than its scan.

**Branch** `sprint-345/p6-bound-and-instruments` · **head** `c313ab1` · cut from
`45aa6c0`. Rollback tag `rollback/pre-345-p6` verified → `45aa6c0` (local
`^{commit}` and the peeled remote ref both). **Model: Opus 5** (`claude-opus-5[1m]`).

```
 src/lib/model-loop/model-call-registry.test.ts |  73 ++++++++++--
 src/lib/model-loop/replay-loop.test.ts         |  87 ++++++++++----
 src/lib/model-loop/replay-loop.ts              |  33 +++++-
 src/lib/model-loop/run-tool-loop.test.ts       | 151 +++++++++++++++++++++++++
 src/lib/model-loop/run-tool-loop.ts            | 101 ++++++++++++++---
 5 files changed, 392 insertions(+), 53 deletions(-)
```

**Blast zone honoured exactly** — five files, all declared. Untouched:
`compare-loop.ts` and `compare-loop.test.ts`, both route handlers,
`scripts/eval-models.mjs` itself, `CLAUDE.md`, `docs/`, `.github/`. No version
bump, no CHANGELOG heading, no publish.

| Commit | Subject |
|---|---|
| `36b8a07` | fix: the tool-result character bound is measured on the assembled body (#355) |
| `adfe82e` | test: the replay timeout case detects the timeout it is named for (#357) |
| `c313ab1` | test: the model-call registry scans beyond src/ and beyond .ts (#356) |

## Criterion 1 — the bound holds on heterogeneous rows

`keepRowsWithinBudget` sized the kept-row count from `rows[0]` alone and never
re-checked the assembled string. It now searches against the **rendered body**
— bare array or whole envelope, framing and all — doubling then bisecting, so
cost tracks rows kept rather than rows received.

Four fixtures, each driven through the real loop with a 50,000-character
budget, each **measured red at `45aa6c0` before the fix** and green after. The
first two are heterogeneous on purpose: a homogeneous fixture is the one shape
where sizing from `rows[0]` happens to be right, which is why the existing
`body.length <= maxChars` assertions were true by construction.

| Fixture | Body fed to the model — RED at `45aa6c0` | GREEN |
|---|---|---|
| **envelope**, sparse first row (62 chars) then 2,999 full rows (569 chars); input 1,709,521 chars | **444,691** (8.9× budget) | **49,681** |
| **bare array**, same shape | **444,664** | **49,654** |
| envelope, three 100 KB rows | **300,082**, marker `showing 3 of 3 rows` with nothing dropped | **26**, marker `showing 0 of 3 rows` |
| envelope, 300 rows pretty-printed | 44,427 **and** `[Truncated: showing 300 of 300 rows]` | 44,427, **no marker**, 300 rows |

RED, verbatim from the run:

```
not ok 5 - #355: a sparse first row does not lift the bound off the envelope branch
  error: 'the body fed to the model must fit the budget: 444691 > 50000'
not ok 6 - #355: the bare-array branch is bounded on heterogeneous rows too
  error: 'the body fed to the model must fit the budget: 444664 > 50000'
not ok 7 - #355: rows too large to fit are dropped, and the marker counts what is there
  error: 'the body fed to the model must fit the budget: 300082 > 50000'
not ok 8 - #355: a result that loses no rows makes no truncation claim
  error: 'nothing was dropped, so nothing may claim it was'
# tests 22 / # pass 18 / # fail 4
```

Every green body is valid JSON with `total_rows` intact and the marker counting
the rows actually present — #331 does not regress, asserted on the wire.

**Note on the issue's own figures.** #355 reports 871,116 characters on a
"sparse first row, rest full" fixture. That figure is **not byte-reproducible**:
the issue does not publish its fixture, and its own numbers imply a
different one (a ~345-char full row and an 18-char first row against my
569/62). The *class* reproduces exactly — with an 18-character first row the
marker arithmetic lands on `showing 2500 of 3000`, the issue's value, verbatim.

## Criterion 2 — the marker tells the truth

Row 4 above. A page that exceeds the budget as it arrives and fits once
re-serialised keeps every row and now carries **no** truncation claim. RED at
`45aa6c0` on the same fixture: `[Truncated: showing 300 of 300 rows]`.

Row 3 is the issue's own case, and its behaviour changes in a second way worth
stating plainly: when a single row will not fit, the model now receives the
envelope with an empty `data`, `total_rows` intact, and `showing 0 of 3 rows`.
The five-row floor this replaces could exceed the budget by any factor. A floor
that breaks the bound is the defect, not a floor.

## Criterion 3 — the replay timeout case detects its own subject

Premise re-verified at `45aa6c0` first: with the whole `Promise.race` block at
`replay-loop.ts:104-119` replaced by `return callTool(name, args)`, **10 pass /
0 fail** — the case named for the race among them.

The case now drives a transport that genuinely never settles.

```
RED  (race deleted, new test):  not ok 9 - a tool call that never settles fails on replay's own timeout, and the run survives
                                  error: 'the run never returned: nothing ended a tool call that never settles'
                                # tests 10 / # pass 9 / # fail 1   (2.15s, process exits cleanly)
GREEN (race restored):          # tests 10 / # pass 10 / # fail 0  (0.27s)
```

The case carries a two-second bound of its own and releases the transport in a
`finally`, so a lost race fails loudly instead of hanging the suite.

`replayLoopOptions` gains a `toolTimeoutMs` seam — 50 ms in the case, which is
why it does not cost 45 seconds. **Production keeps 45 seconds**: the
configuration case asserts `REPLAY_MCP_TOOL_TIMEOUT_MS === 45_000`, and the
source-drift guard, which already asserted the route supplies no tool
transport, now asserts it supplies **neither** seam. Verified directly too —
`replay/route.ts` contains neither `callTool` nor `toolTimeoutMs`.

## Criterion 4 — the registry instrument's claim is true

The scan now covers `src/`, `scripts/` and the four configuration modules at the
repository root — every JS/TS file this repository tracks — across the whole
extension family (`.ts .tsx .mts .cts .js .jsx .mjs .cjs`), written as one
predicate because the gap was an extension the filter had never been asked
about. Paths report relative to the repository root.

`scripts/eval-models.mjs` is registered in **both** maps with its reason:
*carries the full class; migration is the next wave's (#356)*. It belongs in
both because `:432` and `:505` pass `tools` and `:571` does not — verified by
reading the three argument blocks.

RED, with the scan widened and the entry removed — both cases fail, each naming
the file:

```
  error: 'scripts/eval-models.mjs calls chat.completions.create and is not on the registry in
          src/lib/model-loop/model-call-registry.test.ts. Add it with one line saying why it is a
          separate call site — or, better, call runToolLoop.'
  error: 'scripts/eval-models.mjs calls chat.completions.create with tools — a second tool-calling
          loop. This is the shape wave #345 exists to end: call runToolLoop instead.'
# tests 2 / # pass 0 / # fail 2
```

Widening the scan surfaced **no** model caller other than that one — the
complete set is the six already registered plus `scripts/eval-models.mjs`. The
header at `:11-13` now matches what the test checks, and says in its own
paragraph what "a file" means and why getting that wrong is what #356 was.

**The migration is out of scope and was not done.** Every defect #356 lists in
that script is still there.

## Nothing removed

Test *names* compared, `45aa6c0` vs head: **0 names present at base and absent
at head**. 1,055 → 1,059; the four added are the #355 cases. The replay timeout
case kept its exact name, rewritten in place.

## Gates — full output

`npm ci` run first (cold), then:

```
$ npm test
# tests 1092
# suites 10
# pass 1092
# fail 0
# cancelled 0
# skipped 0
# todo 0
# duration_ms 1733.538542
```
(`not ok` lines in the whole TAP stream: **0**. The stream is 1,059 top-level
cases; the summary block above is its complete terminal output.)

```
$ npm run build
FATAL: An unexpected Turbopack error occurred.
Caused by:
- creating new process
- binding to a port
```
Re-measured, as CLAUDE.md requires: this is the sandbox's port-binding
restriction, not the change. `npx next build --webpack` skips the PostCSS
worker pool:

```
$ npx next build --webpack
▲ Next.js 16.3.0 (webpack)
✓ Compiled successfully in 2.9s
  Running TypeScript ...
  Finished TypeScript in 1434ms ...
✓ Generating static pages using 9 workers (32/32) in 179ms
  Collecting build traces ...
Route (app)  … full route table emitted, 32 static pages …
```
CI's `build` job is the real gate and it passed on this branch — Turbopack, on
a runner, green.

```
$ npm run typecheck
> tsc --noEmit
(no output, exit 0)          # run after a successful build
```

```
$ npm run lint
> eslint

/…/scripts/eval-models.mjs
/…/src/components/evidence/AttestationDialog.tsx
/…/src/components/notebook/RenderingCellOutputs.tsx
  96:9  warning  Using `<img>` could result in slower LCP … @next/next/no-img-element
/…/src/lib/bpmn/animation.ts

```
The documented baseline, unchanged — none of the four is in this diff.

```
$ npm run check:compose-env
  PROFILE: db=node-postgres  blob=s3  executor=container  model=openai-compatible
  72 environment entries, 10 build arg(s), 72 applicable spec entr(ies)
  NOTE: 6 variable(s) passed through under a prior-era name. …
```

CI on this PR: `build / test / lint / typecheck` **pass** (1m24s) ·
`container image build` **pass** (1m3s) · `Vercel` **pass** · `DCO` **pass**.
Run: https://github.com/npstorey/civic-ai-tools-website/actions/runs/33176867539

## Item 4 — the runner demonstration: a finding, not a pass

**Anchor criterion 2 is still not demonstrated on a runner, and cannot be by
the procedure this contract specifies.** Reported rather than worked around.

Executed as instructed: branch `throwaway/345-registry-red` cut from this
phase branch, exactly one file added
(`scripts/throwaway-registry-red.mjs` — one `chat.completions.create`, no
`tools`, off the allowlist; placed in `scripts/` as `.mjs` so the same run
would also measure the #356 widening), committed signed, pushed to
`github-https` at `afeeeb0`, **no pull request opened**.

**No CI run was dispatched.** `.github/workflows/ci.yml` triggers on
`pull_request` and `push: branches: [main]` only; `standalone.yml` is
`pull_request`-only; `update-portal-registries.yml` is schedule +
`workflow_dispatch` and does not run the suite. Measured rather than read off
the YAML — polled for 2.5 minutes after the push:

```
$ gh run list --branch throwaway/345-registry-red      # 6 polls, 25s apart
(no output, every poll)

$ gh api ".../actions/runs?branch=throwaway/345-registry-red" --jq .total_count
0

$ gh api ".../commits/afeeeb0.../check-runs"
{"total":1,"runs":[{"name":"Vercel Preview Comments","conclusion":"success"}]}
```

`build / test / lint / typecheck` has no conclusion on that ref because it never
ran. This is precisely the failure mode the wave's own lesson names: **absence
and success look identical**, so it is recorded as absence.

The RED itself is real, and was measured locally on the throwaway commit:

```
  error: 'scripts/throwaway-registry-red.mjs calls chat.completions.create and is not on the
          registry in src/lib/model-loop/model-call-registry.test.ts. …'
# tests 2 / # pass 1 / # fail 1
```

That is a fifth local demonstration, not the runner one the criterion asks for.

**Branch deleted and verified gone:**

```
$ git push github-https --delete throwaway/345-registry-red
 - [deleted]         throwaway/345-registry-red

$ git ls-remote github-https 'refs/heads/throwaway/*'
(empty)
```
Also deleted locally; the commit never reached this phase branch (head is
`c313ab1`, tree clean).

**What would close it** — an owner decision, not mine to take, and both options
are outside this phase's blast zone:

1. add `workflow_dispatch:` (or a `push:` trigger for `throwaway/**`) to
   `.github/workflows/ci.yml`, then re-run this exact procedure; or
2. accept a **draft** PR from the throwaway branch as the dispatch mechanism —
   which this contract forbids ("Open no pull request"), so it needs an explicit
   ruling to reverse.

1. **#355's measured figures are not byte-reproducible.** The issue publishes
   871,116 / 206,740 / 300,113 / 50,055 characters but not the fixture that
   produced them, and its own numbers imply a fixture different from any
   plausible reading (an ~18-char first row with ~345-char rows after it). My
   fixtures measure 444,691 / 444,664 / 300,082. The defect class, the marker
   arithmetic and the order of magnitude all reproduce; the digits do not.
   Nothing turns on it — the criterion is "≤ 50,000", which is met.
2. **Off-by-one in the registry premise.** The contract and #356 both put the
   `.ts`/`.tsx` accept filter at `model-call-registry.test.ts:189`. At
   `45aa6c0` it is at **`:188`**; `:189` is the `.test.tsx?` exclusion. Both
   lines were in the widening, so nothing changes.
3. **`CLAUDE.md`'s Commands table is stale on the test count** — it states
   `npm test` → `# pass 1044`; measured 1088 at `45aa6c0`, 1092 here.
   `CLAUDE.md` is out of zone (and already carries #358's separate correction),
   so it is flagged, not touched.

Everything else held: the rollback tag, `run-tool-loop.ts:393-398`, the stated
contract at `:361-363`, both call sites at `:372`/`:380`, the race block at
`replay-loop.ts:104-119`, the test at `replay-loop.test.ts:357`, `SRC_ROOT` at
`:39`, the header claim at `:11-13`, and `eval-models.mjs`'s three calls at
`:432`/`:505`/`:571` with two passing `tools`.

## Flagged, not fixed — file by file

- `src/lib/model-loop/compare-loop.test.ts:332` — still asserts
  `body.length <= COMPARE_MAX_TOOL_RESULT_CHARS` against a **homogeneous**
  2,000-row fixture, so the assertion remains true by construction. It passes
  unchanged (rows are still dropped on that shape, so the marker is still
  emitted); out of zone, and no change was required.
- `src/lib/model-loop/replay-loop.test.ts:272` — the same true-by-construction
  assertion, in zone. Left alone deliberately: it is #331's case, the #355
  class is now covered at the core where the shared helper actually lives, and
  duplicating a heterogeneous fixture per caller is scope this phase was not
  given. Cheap to strengthen in a follow-up.
- `scripts/eval-models.mjs` — registered, not migrated. Every defect #356 lists
  is still live: the pre-#334 exit at `:557`, raw error text at `:484`,
  `{name, args}`-only records at `:471`, no truncation at all, the superseded
  final-answer wording at `:577`, the unguarded parse at `:466` **outside any
  try**, and portal injection at `:469` without the `name === 'get_data'` guard.
- `docs/deploy.md:583-584` — `MAX_TOOL_RESULT_CHARS` is documented as an
  instance knob but reaches one of three callers (#358/F6), and now also needs a
  line about what an operator sees when one row exceeds the budget (empty
  `data`, `total_rows` intact, `showing 0 of N rows`).
- `.github/workflows/ci.yml` — has no trigger that a pushed non-`main` branch
  can reach. See item 4.
- Out of scope by contract and untouched: #359, #354, #358, #352, #348, #341,
  #342, and #356's migration half.

Closes #355
Closes #357

Refs #356

### DIFF
--- a/src/lib/model-loop/model-call-registry.test.ts
+++ b/src/lib/model-loop/model-call-registry.test.ts
@@ -11,8 +11,19 @@
  * This test says it out loud. Adding a `chat.completions.create` call to a
  * file that is not on the list below fails the suite, and the list carries one
  * line of justification per entry. The wave closed with every tool-calling
- * loop consolidated: the second list below now names exactly one module, and
- * a second entry appearing on it is a regression this test reports by name.
+ * loop in the application consolidated: the second list below names the shared
+ * core and one script, and a further entry appearing on it is a regression
+ * this test reports by name.
+ *
+ * WHAT "A FILE" MEANS HERE, because getting this wrong is what #356 was. The
+ * scan first rooted at `src/` and accepted only `.ts`/`.tsx`, so a fourth
+ * tool-calling loop in `scripts/eval-models.mjs` was invisible to it and the
+ * claim in the paragraph above was false as written, with a live
+ * counter-example in the tree. It now covers `src/`, `scripts/` and the
+ * configuration files at the repository root — every JavaScript or TypeScript
+ * file this repository tracks — in every extension of the family (`.ts`,
+ * `.tsx`, `.mts`, `.cts`, `.js`, `.jsx`, `.mjs`, `.cjs`). A guard whose header
+ * overstates its reach is worse than no guard, because it is trusted.
  *
  * TWO ASSERTIONS, ONE INSTRUMENT. The first is the registry: who calls the
  * model at all. The second is narrower and is the wave's own criterion: how
@@ -36,7 +47,14 @@ import { readdirSync, readFileSync } from 'node:fs';
 import { join, relative, sep } from 'node:path';
 import { fileURLToPath } from 'node:url';
 
-const SRC_ROOT = fileURLToPath(new URL('../..', import.meta.url));
+const REPO_ROOT = fileURLToPath(new URL('../../..', import.meta.url));
+
+/**
+ * Where the scan looks. `src` and `scripts` are walked; the repository root is
+ * read one level deep, which is where the four configuration modules live and
+ * is what keeps `node_modules`, `.next` and the build output out of the walk.
+ */
+const SCAN_DIRECTORIES = ['src', 'scripts'];
 
 /**
  * Every file permitted to call `chat.completions.create`, with the reason it
@@ -56,16 +74,27 @@ const ALLOWED_MODEL_CALLERS: Record<string, string> = {
     'The second copy of that same rubric call, caller-keyed. Out of this wave by ruling D4 (#348).',
   'src/app/api/evidence/generate-summary/route.ts':
     'One summary call, no loop. Out of this wave.',
+  'scripts/eval-models.mjs':
+    'The model-selection harness: a fourth tool-calling loop that carries the full class. Migration is the next wave’s (#356).',
 };
 
 /**
  * Modules carrying a tool-calling loop — a `chat.completions.create` that
- * passes `tools`. Three when this wave opened (`openrouter-streaming.ts`,
- * `evidence/[slug]/replay/route.ts`, `openrouter.ts`); one now. That count is
- * the wave's own first acceptance criterion, and this is where it is measured.
+ * passes `tools`. Three in `src/` when this wave opened
+ * (`openrouter-streaming.ts`, `evidence/[slug]/replay/route.ts`,
+ * `openrouter.ts`); one now. That count is the wave's own first acceptance
+ * criterion, and this is where it is measured.
+ *
+ * The second entry is the debt #356 found, listed rather than fixed: the point
+ * of naming it here is that it stops being invisible to the suite. It is not a
+ * fourth application loop — nothing it writes is served, signed or
+ * user-facing — but it is the same shape, and it chooses which models this
+ * instance offers.
  */
 const ALLOWED_TOOL_LOOPS: Record<string, string> = {
   'src/lib/model-loop/run-tool-loop.ts': 'The shared core.',
+  'scripts/eval-models.mjs':
+    'The model-selection harness: carries the full class; migration is the next wave’s (#356).',
 };
 
 const MODEL_CALL = 'chat.completions.create';
@@ -118,9 +147,9 @@ function passesTools(call: ModelCall): boolean {
 }
 
 function modelCallSites(): ModelCall[] {
-  return sourceFiles(SRC_ROOT).flatMap((path) => {
+  return scannedFiles().flatMap((path) => {
     const source = readFileSync(path, 'utf8');
-    const file = `src/${relative(SRC_ROOT, path).split(sep).join('/')}`;
+    const file = relative(REPO_ROOT, path).split(sep).join('/');
     const calls: ModelCall[] = [];
     let from = 0;
     for (;;) {
@@ -181,12 +210,34 @@ function endOfString(source: string, start: number): number {
   return source.length;
 }
 
+/** Every JavaScript/TypeScript source file in the scanned scope. */
+function scannedFiles(): string[] {
+  return [
+    ...rootLevelFiles(),
+    ...SCAN_DIRECTORIES.flatMap((dir) => sourceFiles(join(REPO_ROOT, dir))),
+  ];
+}
+
+/** The configuration modules at the repository root — read, never walked. */
+function rootLevelFiles(): string[] {
+  return readdirSync(REPO_ROOT, { withFileTypes: true })
+    .filter((entry) => entry.isFile() && isSource(entry.name))
+    .map((entry) => join(REPO_ROOT, entry.name));
+}
+
 function sourceFiles(dir: string): string[] {
   return readdirSync(dir, { withFileTypes: true }).flatMap((entry) => {
     const full = join(dir, entry.name);
     if (entry.isDirectory()) return sourceFiles(full);
-    if (!/\.tsx?$/.test(entry.name)) return [];
-    if (/\.test\.tsx?$/.test(entry.name)) return [];
-    return [full];
+    return isSource(entry.name) ? [full] : [];
   });
+}
+
+/**
+ * A source file of the JavaScript/TypeScript family, tests excluded. Written
+ * as one predicate over the whole family rather than as `.tsx?`, because the
+ * gap #356 found was an extension the filter had never been asked about.
+ */
+function isSource(name: string): boolean {
+  return /\.(c|m)?[jt]sx?$/.test(name) && !/\.test\.(c|m)?[jt]sx?$/.test(name);
 }
--- a/src/lib/model-loop/replay-loop.test.ts
+++ b/src/lib/model-loop/replay-loop.test.ts
@@ -4,10 +4,12 @@
 // is a Next route handler: `node --test` cannot invoke one. So the loop
 // configuration was moved OUT of the route into `replayLoopOptions`, and these
 // tests drive the real `runToolLoop` with the real options factory against a
-// local scripted model server. The only thing substituted is the tool
-// transport — one level BELOW the loop — so no case here restates a cap, a
-// budget, a tool set or the portal injection. Change replay's configuration
-// and these tests change with it; that is the point of the seam.
+// local scripted model server. What a case may substitute is the tool
+// transport — one level BELOW the loop — and, in exactly one case, the
+// per-call timeout, so that a transport which genuinely never settles does not
+// cost 45 real seconds (#357). Neither substitution restates a cap, a budget,
+// a tool set or the portal injection. Change replay's configuration and these
+// tests change with it; that is the point of the seam.
 //
 // A source-drift guard at the bottom closes the remaining gap: it asserts the
 // route actually obtains its options from this factory and supplies no
@@ -38,6 +40,7 @@ import {
   REPLAY_MAX_TOKENS,
   REPLAY_MAX_CUMULATIVE_TOKENS,
   REPLAY_MAX_TOOL_RESULT_CHARS,
+  REPLAY_MCP_TOOL_TIMEOUT_MS,
   type ReplayToolTransport,
 } from './replay-loop.ts';
 import { startScriptedModelServer, type ScriptedReply } from './test-harness.ts';
@@ -96,13 +99,15 @@ interface Wire {
 }
 
 /**
- * Run one replay against a scripted endpoint. `callTool` is the ONLY thing a
- * case supplies beyond the fixtures the route reads off a record — no cap, no
- * budget, no tool list, no portal handling.
+ * Run one replay against a scripted endpoint. `callTool` — and, for the one
+ * case that needs it, `toolTimeoutMs` — are all a case supplies beyond the
+ * fixtures the route reads off a record: no cap, no budget, no tool list, no
+ * portal handling.
  */
 async function runReplay(
   replies: ScriptedReply[],
   callTool: ReplayToolTransport,
+  overrides: { toolTimeoutMs?: number } = {},
 ): Promise<{ payload: ReplayPayload; requests: Record<string, unknown>[] }> {
   const { server, url, requests } = await startScriptedModelServer(replies);
   try {
@@ -115,6 +120,7 @@ async function runReplay(
         systemPrompt: SYSTEM,
         portal: PORTAL,
         callTool,
+        ...overrides,
       }),
     );
     return {
@@ -348,6 +354,11 @@ test('replayLoopOptions carries replay’s own caps and the shared tool set', ()
   assert.equal(options.maxTokens, REPLAY_MAX_TOKENS);
   assert.equal(options.maxCumulativeTokens, REPLAY_MAX_CUMULATIVE_TOKENS);
   assert.equal(options.maxToolResultChars, REPLAY_MAX_TOOL_RESULT_CHARS);
+  assert.equal(
+    REPLAY_MCP_TOOL_TIMEOUT_MS,
+    45_000,
+    'production’s per-tool-call bound: the test seam below overrides it, this is what it overrides',
+  );
   assert.equal(options.finalTurn, 'blocking', 'a route caller has no stream to write into');
   assert.ok(options.tools.length > 0, 'the replay runs against the instance’s MCP tool set');
   assert.equal(options.systemPrompt, SYSTEM);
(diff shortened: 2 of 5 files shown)


npstorey:
**GO — head-bound to `c313ab1`.** Seat read, three-dot against `main` `45aa6c0` (merge-base `45aa6c0`): +392/−53 over the 5 declared files = the PR API; `compare-loop.test.ts` 0 lines. Read in the core: the bound is now measured on the rendered body (`render(kept)` → the string the model receives, doubled then bisected) and falls to raw truncation rather than exceed it. `replay-loop.ts` exports `REPLAY_MCP_TOOL_TIMEOUT_MS = 45_000` with the override injectable and the route passing none (0 hits). The registry scans `['src','scripts']` across every JS/TS extension and carries `scripts/eval-models.mjs` in both maps with the migration named as the next wave's. No `throwaway/*` heads on origin. Runner log for run 33176867539: `# tests 1092 · # pass 1092 · # fail 0` (1088 + 4); my own test-name diff 1003 → 1007, nothing absent at head. Three commits, each sig G with `Signed-off-by` then `Co-Authored-By` contiguous. Four required checks SUCCESS on this exact head; `CLEAN`. Closes #355, #357; the registry half of #356.

**On item 4:** verified at `main` — `ci.yml` triggers on `pull_request` and `push: branches: [main]` only, so the G2 procedure could not dispatch a run by construction. **Seat recommendation to the owner: the draft-PR demonstration, yes** — a draft PR from a throwaway branch carrying one violating file, the registry job goes red on a runner, PR closed, branch deleted; no workflow edit, so the gate is tested without being changed. Run it after this merge, before the closeout PR. Owner's key on the merge: merge commit. A moved head voids this GO.

PR #362, opened 2026-08-28
docs: wave N7 close — the loop rule rewritten to the tree, and a knob's real scope (#345)

npstorey:
Wave N7 (#345) closeout. **Docs only — no source file changes.**

## The three corrections

**1. `CLAUDE.md`'s loop rule described a tree that no longer exists.**

It sent a reader to `replay/route.ts` — which now contains **zero**
`chat.completions.create` — and to `openrouter.ts`, which has one, no-tools, in
`queryWithoutMcp`. It named neither `src/lib/model-loop/run-tool-loop.ts` nor
the registry test that supersedes its grep. And the grep it prescribed,
`grep -rl "chat.completions.create" src`, is scoped to `src` — the exact scope
that hid a fourth loop in `scripts/eval-models.mjs` (#356).

Rewritten to the tree as it is: one core, factories beside it, and the
bidirectional registry test named as the guard to run instead of a grep. The
incident comment keeps N6's class lesson and adds N7's, which is the same lesson
one level up — *a check scoped to a directory cannot see a defect scoped to a
behaviour.*

**2. `MAX_TOOL_RESULT_CHARS` was documented without its scope.**

`docs/deploy.md` listed it among the instance tuning knobs with no qualifier. It
is read on the streaming query path only (`openrouter-streaming.ts`); the replay
and `/api/compare` callers pass a fixed `50_000`. Setting it on an instance today
does nothing to two of the three callers. Now stated where an operator reads it
(#358).

**3. The commands table claimed `# pass 1044`.**

Measured **1092** on the runner at `859abae` —
[run 33178112116](https://github.com/npstorey/civic-ai-tools-website/actions/runs/33178112116),
`# tests 1092 / # pass 1092 / # fail 0`. Taken from CI rather than a local run,
because that table is what a reader checks their own run against.

## The retrospective entry

Beyond the three named corrections, and flagged as such: `docs/RETROSPECTIVE.md`
is where this repo keeps session retros, and N6's closeout set the precedent. The
entry records what the wave did, four things the phases got right that were not
asked for, what the cold read found — **including a defect the wave itself
introduced while closing #331** — the three ORCH-layer premise failures the
phases caught, and what the wave deliberately did **not** do, so the next census
does not rediscover those as findings.

Strike that section if the closeout was meant to be the three corrections alone.

## Verification

`npm run lint` → `✖ 4 problems (0 errors, 4 warnings)`, the documented baseline,
unchanged. No source file is touched, so no behaviour changes.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -29,7 +29,7 @@ remote) — don't design an owner-run leg around the assumption that it doesn't.
 
 | Command | Healthy output |
 |---|---|
-| `npm test` | `# pass 1044` / `# fail 0` (`node --test` TAP summary) |
+| `npm test` | `# pass 1092` / `# fail 0` (`node --test` TAP summary) |
 | `npm run build` | `✓ Compiled successfully`, then the Route (app) table |
 | `npm run typecheck` | no output, exit 0 — **run after `npm run build`** (tsconfig includes `.next/types`, which the build emits) |
 | `npm run lint` | `✖ 4 problems (0 errors, 4 warnings)` — warnings are the baseline; **zero errors** is the gate |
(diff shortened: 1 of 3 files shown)


npstorey:
**GO — head-bound to `5077605`.** Seat read, three-dot against `main` `859abae` (merge-base `859abae`): +56/−9 over 3 files = the PR API, no `src/` file touched; one commit, sig G, both trailers contiguous; four required checks SUCCESS on this exact head; `CLEAN`. Claims read against the tree: `# pass 1092` matches run 33178112116 on `main` at `859abae`; the rewritten loop rule is true at this head — one core, the two factories, the bidirectional registry test, and its scan of `src/`, `scripts/` and the root-level modules (`rootLevelFiles()` at the test's `:216`); the `MAX_TOOL_RESULT_CHARS` qualification matches the fixed 50,000 the replay and compare factories pass. The retrospective entry stays — it is this repo's orchestration log, and N6's closeout set the pattern. Closes #358. Owner's key: merge commit. A moved head voids this GO.

PR #364, opened 2026-08-28
fix: two surfaces stop asserting numbers they did not compute (#339, #307)

### LINKED ISSUE #339: buildStatsSummary counts catalog-search hits as "records analyzed" in reader-facing and published output
## Summary

The "N records analyzed" line shown to readers and carried into published records now counts catalog-search hits as records. It was dormant for one data source and became live when result summaries began populating for it.

## Measured at `1ddd61a`

`buildStatsSummary` (`src/lib/streaming.ts:1083`):

```ts
const totalRows = toolsCalled.reduce((sum, t) => sum + (t.resultSummary?.rows || 0), 0);
if (totalRows > 0) statParts.push(`${totalRows.toLocaleString()} records analyzed`);
```

The reduce runs over **all** tool calls with no operation-type filter. A catalog search returning 40 dataset descriptions contributes 40 to "records analyzed", alongside rows actually retrieved from a dataset.

`buildProvenanceLine` (`:1150`) does not have this problem — it filters to query calls first. The two lines can therefore disagree about the same run, and the one that overcounts is the more prominent.

## Why this is evidence integrity, not cosmetics

This string is reader-facing on the primary answer surface and describes how much data backed a published claim. "Records analyzed" naming search hits inflates the apparent evidentiary weight of an answer, and it does so silently.

It was previously unreachable for one source because result summaries were always absent there, so no prior run could have surfaced it.

## Related, same root

`resultRows`/`resultColumns` now appear in signed packages' `queries[]` (`src/lib/evidence/packager.ts:394-395`) and `tool.response_rows` on trace spans where both were previously always absent for that source. Both are honest and correct — noted because they are new content in published records that no issue asked for.

## Suggested direction

Filter to query-type calls, matching `buildProvenanceLine`, or name the line for what it counts.

npstorey:
Wave N8, phase **P4a** (lane 2) of sprint #363. Closes #339 and #307.

Two reader-facing surfaces asserted numbers they had not computed. Both are attestation surfaces: the first is carried into published record packages, the second is the public record index.

- **#339** — `buildStatsSummary` summed `resultSummary.rows` over *every* tool call with no operation-type filter and rendered the total as `N records analyzed`. A catalog search returning 40 dataset **descriptions** contributed 40 beside a query returning 12 actual rows, so the most prominent number on the answer surface read **52** while the provenance line under it (which already filtered to query calls) read **12**.
- **#307** — `GET /api/evidence/list` ran the attestation-count query only when `sort=attested`, then emitted `attestationCount: … || 0` under every sort. Measured against production: every record on the public list reported `0` while the database held nine attestation rows across seven records.

Both are `docs/design-principles.md` principle 3 (no false precision) failing in the direction that matters — a specific figure, rendered confidently, that overstates.

## Branch and diff

Branch `sprint-363/p4a-signed-counts`, from `96c4f76` (tag `rollback/pre-n8-p4a`). Two commits.

```
 src/app/api/evidence/list/route.ts   |  45 +++---
 src/lib/evidence/record-list.test.ts | 265 +++++++++++++++++++++++++++++++++++
 src/lib/evidence/record-list.ts      | 141 +++++++++++++++++++
 src/lib/streaming-stats.test.ts      | 215 ++++++++++++++++++++++++++++
 src/lib/streaming.ts                 |  55 +++++++-
 5 files changed, 700 insertions(+), 21 deletions(-)
```

**Blast zone honoured — five files, every one declared.** `src/lib/openrouter-streaming.test.ts`, `src/lib/model-loop/*`, `src/lib/openrouter-streaming.ts`, `src/app/api/compare-stream/route.ts` and `src/app/api/query-notebook/route.ts` — lane 1's files — are untouched (`git diff --name-status 96c4f76..HEAD` lists exactly the five above). `ProgressLog.tsx`, `McpResponseDisplay.tsx` and `dashboard/page.tsx` were read for confirmation and not edited.

## The lane prediction held

ORCH predicted at G2 that the query-only filter would **not** move `openrouter-streaming.test.ts:504` (`assert.match(stats, /\b5 records analyzed\b/)`), because both tool calls in that fixture carry `type: 'query'` at `:447-448`.

**Held.** `git diff --exit-code -- src/lib/openrouter-streaming.test.ts` is clean, the file is byte-identical to the base, and the test passes unmodified: `ok 9 - queryWithMcpStreaming: buildStatsSummary and buildProvenanceLine sum data.length (5) across calls, never total_rows (300) and never 0`.

## What changed

**#339 — one predicate, three readers.** `isQueryCall` (exported) and `sumQueryRows` in `src/lib/streaming.ts`. The volume count, the query count and the provenance line all read the same predicate, so they cannot drift apart again. `buildProvenanceLine` moves onto it too — it read `operationType` alone, so a call recorded with only `args.type` (a legacy package read back) produced a records count with no `Source:` line beneath it; it now produces both or neither.

**#307 — the count is computed on every listing, and the projection is extracted.** The query runs whenever the page has rows. The row→response projection moves to `src/lib/evidence/record-list.ts`, in the shape `src/lib/evidence/readback.ts` already uses, because the route imports `db` at module scope and the test runner (no path aliases, cannot load a route module) could not reach it. `attestationCounts` is `Map | null`: a Map means computed, so a record missing from it genuinely has none and `0` is true; `null` means not computed and the field is **omitted**. A caller that did not count has no way, through this function, to say "zero" instead of saying nothing.

## Acceptance criteria

| # | Criterion | Evidence |
|---|---|---|
| 1 | `records analyzed` counts query rows only | `#339: a catalog search returning 40 hits beside a query returning 12 rows renders "12 records analyzed"` |
| 2 | `buildStatsSummary` and `buildProvenanceLine` agree | `#339: the two lines report the SAME row count for the same run`, plus an exhaustive sweep over every pair of operation labels |
| 3 | The index computes what it asserts | `#307: a record with two attestations lists 'attestationCount: 2'` + `#307: the attestation-count query in the list route is not gated on the sort` |
| 4 | Never asserted when not computed | `#307: an UNCOMPUTED count is ABSENT, never '0'` and `#307: a COMPUTED zero is stated — the two zeros are different facts` |
| 5 | Pinned by a test that runs without a database | `src/lib/evidence/record-list.test.ts` — fixture rows, schema imported for types only, no `db` |
| 6 | Nothing lost | 1092 → **1111**, `# fail 0`, zero test names absent (set difference over the TAP names, both runs) |
| 7 | Trailers | both commits carry `Signed-off-by` then `Co-Authored-By`, one contiguous block, author matching |

**RED demonstrated before green, three separate injections:**

- `sumQueryRows` reverted to the base unfiltered reduce → all 8 `#339` tests fail; the first renders `52 records analyzed · 1 query · 3.2s`.
- the base `list/route.ts` restored → the two route-source guards fail.
- the projection reverted to the base's inline `|| 0` → the two uncomputed-count assertions fail.

## Gates

`npm ci` first. `npm test` 1111/1111; `npm run build` compiled, route table emitted; `npm run typecheck` exit 0, no output; `npm run lint` `✖ 4 problems (0 errors, 4 warnings)` — the documented baseline, unchanged; `npm run check:compose-env` `RESULT: PASS`.

Model: Opus 5.

Draft — ORCH merges on evidence-pass. Full phase report, including flagged-not-fixed findings and the contract premises that did not survive, is in the phase report to ORCH.

### DIFF
--- a/src/app/api/evidence/list/route.ts
+++ b/src/app/api/evidence/list/route.ts
@@ -3,6 +3,7 @@ import { db } from '@/lib/db';
 import { evidenceRecords, attestationPackages, users } from '@/lib/db/schema';
 import { eq, desc, asc, ilike, or, and, gte, isNull, isNotNull, sql } from 'drizzle-orm';
 import { visibilityMatches } from '@/lib/evidence/visibility-sql';
+import { buildRecordListItems, sortByAttestationCount } from '@/lib/evidence/record-list';
 
 const PAGE_SIZE = 20;
 
@@ -122,9 +123,24 @@ export async function GET(request: NextRequest) {
     : [];
   const creatorMap = new Map(creators.map(c => [c.id, c]));
 
-  // Fetch attestation counts if sorting by 'attested'
-  let attestationCountMap = new Map<string, number>();
-  if (sort === 'attested' && records.length > 0) {
+  // Fetch attestation counts — on EVERY listing, not only `sort=attested`
+  // (#307).
+  //
+  // The query used to be gated on the sort, while the projection below emitted
+  // `attestationCount` unconditionally. Under every other sort the map was
+  // empty and the field fell through to `0`, so the public index told every
+  // reader that no one had reviewed any record — measured against production,
+  // where the database held nine attestation rows across seven records.
+  //
+  // The count is a review signal on a page whose whole job is scrutiny, so a
+  // false zero is worse than a slower page. It is one grouped count over at
+  // most `PAGE_SIZE` ids, on a page that already runs three queries.
+  //
+  // `null` (rather than an empty map) is what the projection reads as "not
+  // computed" and omits the field for; it is never `0`. See
+  // `@/lib/evidence/record-list`.
+  let attestationCountMap: Map<string, number> | null = null;
+  if (records.length > 0) {
     const counts = await db
       .select({
         evidenceRecordId: attestationPackages.evidenceRecordId,
@@ -141,23 +157,18 @@ export async function GET(request: NextRequest) {
     attestationCountMap = new Map(counts.map(c => [c.evidenceRecordId, c.count]));
   }
 
-  const results = records.map(r => ({
-    id: r.id,
-    slug: r.slug,
-    title: r.title,
-    summary: r.summary.slice(0, 200) + (r.summary.length > 200 ? '...' : ''),
-    model: r.model,
-    verificationStatus: r.verificationStatus,
-    withdrawnAt: r.withdrawnAt?.toISOString() || null,
-    reinstatedAt: r.reinstatedAt?.toISOString() || null,
-    createdAt: r.createdAt.toISOString(),
-    creatorName: creatorMap.get(r.creatorId)?.displayName || 'Unknown',
-    attestationCount: attestationCountMap.get(r.id) || 0,
-  }));
+  // The row -> response projection lives in `@/lib/evidence/record-list` so it
+  // is reachable by the test runner, which cannot load this module (it imports
+  // `db` at module scope). Same reason `[slug]`'s read-back projection lives in
+  // `@/lib/evidence/readback`.
+  const results = buildRecordListItems(records, {
+    creators: creatorMap,
+    attestationCounts: attestationCountMap,
+  });
 
   // Re-sort by attestation count if needed
   if (sort === 'attested') {
-    results.sort((a, b) => b.attestationCount - a.attestationCount);
+    sortByAttestationCount(results);
   }
 
   return NextResponse.json({ records: results, total, page, totalPages });
--- a/src/lib/evidence/record-list.test.ts
+++ b/src/lib/evidence/record-list.test.ts
@@ -1,0 +1,265 @@
+// Tests for the `GET /api/evidence/list` row → response projection
+// (record-list.ts) and for the property #307 turns on: the public index never
+// states an attestation count it did not compute.
+//
+// WHY THIS FILE EXISTS
+//
+// The route ran the attestation-count query only when `sort=attested`, then
+// emitted `attestationCount: map.get(id) || 0` under every sort. Measured
+// against production, every record on the public list reported `0` while the
+// database held nine attestation rows across seven records. A count reading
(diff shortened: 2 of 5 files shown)


npstorey:
**Seat GO — head-bound to `4bc21c29ecf21fe087c3af82deefae4cd497304e` (`4bc21c2`).** A moved head voids this.

- `git diff 96c4f76...4bc21c2 --numstat`: 5 files, +700 −21 — equal to the PR API's `additions`/`deletions`/`changedFiles`; the five paths are the declared zone and nothing else.
- Check rollup on the exact head: `build / test / lint / typecheck` success, `container image build` success, `DCO` success, `Vercel` success (run 33191514273). All four required checks, enumerated on the commit, not on the branch.
- Both commits (`22bc9e1`, `4bc21c2`) signed (`G`); trailers `Signed-off-by` then `Co-Authored-By: Claude Opus 5 (1M context)`, contiguous and in order.
- The load-bearing claims, read in the diff: `isQueryCall` is the one predicate (`operationType || args.type === 'query'`) and both reader-facing lines sum through `sumQueryRows`; the index runs the attestation count on every listing, passes `null` when it did not compute, and `buildRecordListItem` omits the field on `null` and states a computed `0` — the honest-absence shape, not an asserted default. Tests pin both REDs: 40 catalog hits beside 12 query rows renders `12 records analyzed`; two attestations list `attestationCount: 2`; an uncomputed count is absent, never `0`.
- The disclosed behaviour change (`buildProvenanceLine` now emits a `Source:` line for a legacy record carrying only `args.type`) is a render-time formatter on the same predicate; no signed bytes change. Accepted.

**Sequencing, per the ORCH's recommendation and the owner's relay: the merge waits until P2 reports.** Lane 1 lands first; P4a merges after with no rebase (the files are disjoint; `strict_required_status_checks_policy` is `false`). The PR is a **draft** — mark it ready for review before the merge button; the GO stays bound to `4bc21c2` across that state change because the head does not move.

Prediction accepted onto the record: P4a alone lands 1111; after P2, 1111 + P2's additions with no name absent from either set, read from the runner's job log before the second merge.

Flagged items 1–4 in the G3 record are noted; item 1 (the `users` table fetched on every listing) is worth its own issue — file it.

PR #365, opened 2026-08-28
fix: portal injection and the tool bound become the loop's, not each caller's (#359, #352)

### LINKED ISSUE #359: A tool call's span and its recorded args disagree about the portal, inside the signed trace

npstorey:
Wave N8, phase P2 (lane 1). Base `96c4f76`. Model: **Opus 5**. Closes #359, #352.

## What changed

`runToolLoop` gains `portal` and `toolTimeoutMs`. Injection runs **above** the tool-call record, the `tool_start` emit and the `mcp_tool_call` span; the timer is raced and cleared in one `finally` in the core. The four in-`src` caller copies are gone, and `/api/compare` gets a per-tool-call bound it has never had.

| File | Change |
|---|---|
| `src/lib/model-loop/run-tool-loop.ts` | two options; the guard above `deriveOperationType`; `boundToolCall` races and clears once |
| `src/lib/model-loop/compare-loop.ts` | closure deleted; passes `portal` + `COMPARE_MCP_TOOL_TIMEOUT_MS` (45s, new) |
| `src/lib/model-loop/replay-loop.ts` | closure deleted; passes `portal` + `toolTimeoutMs` |
| `src/lib/openrouter-streaming.ts` | `queryWithMcpStreaming` grows a trailing `toolCallOptions` and forwards both |
| `src/app/api/compare-stream/route.ts` | transport only; forwards `{ portal, toolTimeoutMs }` |
| `src/app/api/query-notebook/route.ts` | transport only; forwards `{ portal, toolTimeoutMs }` |
| `src/lib/model-loop/injection-and-bound.test.ts` | new — 18 cases across all four callers |

## Why the span and the record disagreed (#359)

The closure runs at `run-tool-loop.ts:653` — after the record at `:630`, after `tool_start` at `:633`, thirteen lines after the span stringified `args` at `:640-647`. So on exactly the calls a portal was injected into, the span carried no `tool.portal_domain`, while `packager.ts` and the replay identity key read the mutated object and did carry it.

Two deliberate details in the guard:

- **`!argumentsMalformed` is new and load-bearing.** A caller's closure never ran on an unparseable argument set, because the core throws first. Injection at this height does run, on the `{}` the failed parse left behind — without the clause, a call whose arguments never parsed would be recorded, spanned and **signed** carrying a portal it never had.
- **`name === 'get_data'` stays literal** rather than widening to a source lookup: `search` and `fetch` have `additionalProperties: false` schemas with no portal property, so an injected portal would be stripped upstream and still corrupt the recorded arguments.

A latent divergence is **named in a comment beside the guard, not fixed**: the data source resolves `domain || portal` with `domain` winning, and the guard looks only at `portal`. It cannot fire while `get_data` advertises no `domain` property. The comment names the trigger for whoever adds one.

## Acceptance

| # | Criterion | Result |
|---|---|---|
| 1 | One injection in `src/` | `git grep -n -P '(args\|toolArgs)\.portal\s*=[^=]'` (all tracked files, no pathspec) → 2 non-test sites: the core, and `scripts/eval-models.mjs:469` (a later phase's) |
| 2 | The span agrees with the record, on every caller | 4 cases: the core, `compare-loop`, `replay-loop`, `queryWithMcpStreaming` |
| 3 | A malformed call is never injected into | record `{}`, span `tool.arguments` `'{}'`, no `tool.portal_domain`, still fails as `unknown` |
| 4 | No timer outlives its call | 6 cases under `mock.timers` + a timer ledger; the bound fires on a hang on all three bounded callers |
| 5 | Nothing lost | `# tests 1110 / # pass 1110 / # fail 0`; 0 base test names absent, 18 added |
| 6 | Trailers | `Signed-off-by` then `Co-Authored-By`, one contiguous block |

## Every criterion shown red before it was accepted

Four probes, each applied to the merged code and reverted:

1. **Drop `!argumentsMalformed`** → criterion 3 fails: `{} !== { portal: 'data.cityofnewyork.us' }`.
2. **Move the injection back below the record and the span** → 5 cases fail, including `queryWithMcpStreaming`, with the exact #359 shape: span args missing `portal`, record args carrying it.
3. **Delete the `finally { clearTimeout }`** → 2 cases fail: `1 !== 0` pending timers.
4. **Delete the race entirely** → 5 cases fail; the hang cases report `'hung' !== 'settled'` after the suite's own 2s real-clock guard.

## Checks

`npm ci` first. `npm test` 1110/1110. `npm run build` compiled successfully. `npm run typecheck` exit 0, no output. `npm run lint` `✖ 4 problems (0 errors, 4 warnings)` — the baseline. `npm run check:compose-env` PASS.

**Draft.** Merge is the ORCH's on evidence-pass; this branch pushes no tag and never touches `main`.

### DIFF
--- a/src/app/api/compare-stream/route.ts
+++ b/src/app/api/compare-stream/route.ts
@@ -205,33 +205,26 @@ Be honest if you don't have access to current or real-time data.`;
     // Run queries (both in parallel, or MCP-only when mcpOnly flag is set)
     const runQueries = async () => {
       try {
-        const MCP_TOOL_TIMEOUT_MS = 45_000; // 45s timeout per individual tool call
+        // 45s per individual tool call. The route states the value; the race
+        // and the timer's disposal live in the loop core (#352) — this file
+        // used to arm a timer inside the tool closure and never clear it, so a
+        // call that answered in 200 ms left one pending for the rest of the
+        // bound.
+        const MCP_TOOL_TIMEOUT_MS = 45_000;
 
         const mcpQuery = queryWithMcpStreaming(
           query,
           model,
           mcpTools,
-          async (name, args) => {
-            // Socrata tools expect a portal; Data Commons tools don't.
-            if (name === 'get_data' && !args.portal) {
-              args.portal = portal;
-            }
-            // Race the MCP call against a timeout so one slow tool call
-            // cannot hang the entire SSE stream indefinitely.
-            const result = await Promise.race([
-              callMcpTool(name, args),
-              new Promise<never>((_, reject) =>
-                setTimeout(
-                  () => reject(new Error(`MCP tool "${name}" timed out after ${MCP_TOOL_TIMEOUT_MS / 1000}s`)),
-                  MCP_TOOL_TIMEOUT_MS,
-                )
-              ),
-            ]);
-            return result;
-          },
+          // Just the transport. Portal injection moved into the core (#359):
+          // done here, it ran after the core had recorded the call and
+          // stringified its arguments onto the span, so the span reported no
+          // portal on exactly the calls that had just been given one.
+          callMcpTool,
           systemPromptWithMcp,
           callbacks,
           { builder: trace, parentSpanId: trace.rootSpanId, systemPromptHash, resolveToolSource: (name) => routeTool(name).sourceId },
+          { portal, toolTimeoutMs: MCP_TOOL_TIMEOUT_MS },
         );
 
         if (mcpOnly) {
--- a/src/app/api/query-notebook/route.ts
+++ b/src/app/api/query-notebook/route.ts
@@ -379,6 +379,8 @@ async function runPhaseA(args: {
   return new Promise<CompletionResult>((resolve, reject) => {
     let completionResult: CompletionResult | null = null;
     const sentToolCalls = new Set<string>();
+    // 45s per individual tool call. The route states the value; the race and
+    // the timer's disposal live in the loop core (#352).
     const MCP_TOOL_TIMEOUT_MS = 45_000;
 
     const callbacks: StreamCallbacks = {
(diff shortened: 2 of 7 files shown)


npstorey:
**Seat GO — head-bound to `3e189f97fb6dc4152d0b24cc50e9f8bcc405b18b` (`3e189f9`).** A moved head voids this.

- `git diff 96c4f76...3e189f9 --numstat`: 7 files, +1059 −93 — equal to the PR API's `additions`/`deletions`/`changedFiles`; six declared files plus the new `injection-and-bound.test.ts`, nothing else. P4a's five files: empty three-dot diff on this branch.
- Check rollup on the exact head: `build / test / lint / typecheck`, `container image build`, `DCO`, `Vercel` — all success. Enumerated on the commit.
- One commit, signed (`G`); `Signed-off-by` then `Co-Authored-By: Claude Opus 5 (1M context)`, contiguous, in order.
- The structural claim, read in the blob at `3e189f9` rather than the record: injection at `run-tool-loop.ts:740` (`portal && !argumentsMalformed && name === 'get_data' && !args.portal`), then the record at `:746`, the `tool_start` event at `:749`, `tool.arguments` at `:759` and `tool.portal_domain` at `:762`, the malformed-args throw at `:766`, the bound call at `:769`. `boundToolCall` clears its timer in a `finally`. #359 closes by construction.
- Injection census at this head over all tracked files, no pathspec: exactly two non-test sites — the core at `:740` and `scripts/eval-models.mjs:469`, which P3 removes. Criterion 1 reaches "one" at P3, as chartered.
- No `setTimeout` race remains in either streaming route; all four callers pass `toolTimeoutMs` (`compare-loop.ts:128` per D2-A, `replay-loop.ts:112/135`, the two routes at `:227`/`:443`, streaming via its options at `:276`). #352 closes once.
- Eighteen tests in `injection-and-bound.test.ts` pin it, including the property-shaped ones: the span agrees with the record on every caller, a malformed call is never given a portal, and "no caller in `src/` carries an injection or a tool timer of its own".

**Sequence, as the G4 record states it:** mark #365 ready → merge P2 → `rollback/pre-n8-p2-merged` → merge P4a on its standing GO (`4bc21c2`, PR #364) → `rollback/pre-n8-p4a-merged` → read the merged head's `push` run from the job-log API.

**The prediction is a stop condition, accepted onto the record:** merged head `# tests 1129`, `# fail 0`, no name absent from either branch's set. Any other number means the lanes interacted; stop and report, do not explain.

P4b's contract re-anchors at the merged head (all six sites moved under this diff). Flagged item 3 (`/api/compare` now has a 45s ceiling and `docs/instance-setup.md` §5.3 does not say so) joins P6's docs item beside #354.

PR #367, opened 2026-08-28
docs: write down how a fix arrives — issue, measurement, branch, merge

npstorey:
## Summary
- Adds a "How a fix arrives" section to CONTRIBUTING.md: file an issue carrying a measurement (what ran, what happened, at which SHA), lead the report with what a reader of the output would wrongly believe (a false record is quieter than a failure, and more serious for a record-publishing app), branch off current `main` rather than cherry-picking a reference fix from elsewhere, and the merge rule as it stands — protected `main`, no direct pushes, PR required, merge deploys to production.
- Docs only; no application code or test changes.

## Test plan
- [x] Read-through against the current CONTRIBUTING.md to confirm no duplication with the existing "Commits, signing, and how we merge" section
- N/A — docs-only change, no build/test/lint required

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -24,6 +24,36 @@ Thank you for your interest in contributing! This is the demo website for the ci
 4. Create a branch for your changes
 5. Submit a pull request
 
+## How a fix arrives
+
+"Getting started" above covers cloning and branching for any contribution. For a bug fix
+specifically, here's the fuller path from issue to production:
+
+1. **Open an issue, and bring a measurement.** State what you ran, what happened, and the commit
+   SHA it reproduces at — for example, "`npm test` at `a1b2c3d` fails with `expected 1092, got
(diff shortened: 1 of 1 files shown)


npstorey:
Seat GO - head-bound to 78e966e4f28c49e74931f1964774bf7797d13a90 (78e966e). A moved head voids this. Verified by the seat on 2026-08-28 from the remote ref and the runner: three-dot diff +30/-0 on CONTRIBUTING.md alone, equal to the PR API; all four required checks plus Vercel green on this exact commit; one signed commit with Signed-off-by then Co-Authored-By naming the model, contiguous; the added text is neutral (no organization, deployment or engagement named), extends the #335 section rather than duplicating it, and point 4 now names the four required checks. Docs-only and outside every Wave N8 blast zone, so it merges between N8 merges with no coordination beyond notice to ORCH 363. Merge-commit only.

PR #368, opened 2026-08-28
RED PROBE (do not merge) — N8 P3 criterion 3: a model call under a new top-level directory

npstorey:
**Wave N8 P3 — acceptance criterion 3. This PR is a red demonstration and is closed unmerged. Do not merge.**

It carries the P3 branch plus one extra file, `tools/x.mjs`: a `chat.completions.create` that passes `tools`, under a top-level directory this repository does not have.

At `cecc959` the registry guard's scan was `SCAN_DIRECTORIES = ['src', 'scripts']` plus a one-level, non-recursive read of the repository root. A file at `tools/x.mjs` sits in none of those, so the guard would have stayed green — confirmed by execution locally, not inferred: a probe reproducing that scan covered 261 files and did not include `tools/x.mjs`.

With the scan derived from `git ls-files`, both assertions in `src/lib/model-loop/model-call-registry.test.ts` report it by name. The expected failure in `build / test / lint / typecheck` is:

```
tools/x.mjs calls chat.completions.create and is not on the registry in src/lib/model-loop/model-call-registry.test.ts.
tools/x.mjs calls chat.completions.create with tools — a second tool-calling loop.
```

The phase branch itself is `sprint-363/p3-fourth-loop-onto-core`; `tools/x.mjs` does not exist on it.

### DIFF
--- a/scripts/eval-models.mjs
+++ b/scripts/eval-models.mjs
@@ -3,25 +3,54 @@
 /**
  * Model evaluation harness for civic data queries.
  *
- * Runs a set of test queries against multiple LLMs via OpenRouter,
- * using the Socrata MCP server for tool calls. Captures raw responses,
- * token usage, latency, and tools called for manual scoring.
+ * Runs a set of test queries against multiple LLMs via OpenRouter, using the
+ * Socrata MCP server for tool calls. Captures raw responses, token usage,
+ * latency, and tools called for manual scoring.
+ *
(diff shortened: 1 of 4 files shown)


npstorey:
Red demonstration complete and recorded. Run 33196411536, check `build / test / lint / typecheck`, job 98934572252 — failed with both registry assertions naming tools/x.mjs, and nothing else in the suite (1131 tests, 2 fail). Closing unmerged; the file does not exist on sprint-363/p3-fourth-loop-onto-core.

PR #369, opened 2026-08-28
fix: the fourth tool-calling loop drives the shared core, and the guard's scope stops being a list (#356)

### LINKED ISSUE #356: A fourth tool loop lives in scripts/eval-models.mjs, where neither the census nor the registry test can see it

npstorey:
**Wave N8 — Phase P3 (lane 1). Closes #356.** Base `cecc959`. Model: **Opus 5**. Draft.

## What changed

`scripts/eval-models.mjs` drives `runToolLoop`. Its own loop, its unguarded argument parse, its raw-error-to-the-model path and its portal injection are **deleted, not wrapped**. `ALLOWED_TOOL_LOOPS` drops to one entry and `ALLOWED_MODEL_CALLERS` loses the harness in the same commit. The registry guard's universe is now derived from `git ls-files` rather than a two-directory list, and its header says exactly what the scan does.

## Blast zone

Exactly the three declared files, no more:

```
 scripts/eval-models.mjs                        | 322 ++++++++-----------------
 scripts/eval-models.test.mjs                   |  81 ++++++-
 src/lib/model-loop/model-call-registry.test.ts | 139 ++++++-----
 3 files changed, 253 insertions(+), 289 deletions(-)
```

`src/lib/model-loop/run-tool-loop.ts` is **untouched** — this phase is a consumer of the core, not an editor of it. `package.json` needed no change (there is no `eval-models` script; P1 measured this and the chartered zone was wrong). The throwaway `tools/x.mjs` does not exist on this branch.

## Acceptance

| # | Criterion | Evidence |
|---|---|---|
| 1 | One loop | `ALLOWED_TOOL_LOOPS` has one entry and the bidirectional assertion passes, so the tool-passing set is exactly `{run-tool-loop.ts}` across every tracked file |
| 2 | One injection, at last | `git grep -n -P '(args\|toolArgs)\.portal\s*=[^=]'` over all tracked files, no pathspec → **one** non-test site, `run-tool-loop.ts:740`. The wave's criterion 1 |
| 3 | Scope true by construction, red **on a runner** | Draft PR #368, run [`33196411536`](https://github.com/npstorey/civic-ai-tools-website/actions/runs/33196411536), check **`build / test / lint / typecheck`**, job `98934572252` — **fail**, both assertions naming `tools/x.mjs`, `# tests 1131 / # fail 2`. Closed unmerged, branch deleted |
| 4 | Bidirectional | Four probes shown red — a removed real entry and a stale entry, on each registry |
| 5 | Header states what the scan does | Rewritten, including the two consequences a reader would otherwise infer |
| 6 | Nothing lost | 1129 → **1131**, `# fail 0`; TAP name sets diffed, absent set **empty**, added set exactly the two new tests |
| 7 | Trailers | Both commits: `Signed-off-by` then `Co-Authored-By`, contiguous |

## One forced behaviour change, disclosed

The harness's budget guardrail had three tiers it implemented itself: a soft cap, a hard cap at 2× that returned a synthetic `[HARD ABORT]` string with no summary turn, and a per-round projection. The core carries one bound. The soft tier survives exactly; **`hardAbort` is removed from the result** rather than written out as a field that can no longer be true. Stated in the file beside the constants, and flagged for the gate.

### DIFF
--- a/scripts/eval-models.mjs
+++ b/scripts/eval-models.mjs
@@ -3,25 +3,54 @@
 /**
  * Model evaluation harness for civic data queries.
  *
- * Runs a set of test queries against multiple LLMs via OpenRouter,
- * using the Socrata MCP server for tool calls. Captures raw responses,
- * token usage, latency, and tools called for manual scoring.
+ * Runs a set of test queries against multiple LLMs via OpenRouter, using the
+ * Socrata MCP server for tool calls. Captures raw responses, token usage,
+ * latency, and tools called for manual scoring.
+ *
+ * IT DRIVES THE SHARED TOOL-CALLING LOOP (#356). Until this file was migrated
+ * it carried a fourth copy of that loop: its own iteration bound, its own
+ * unguarded `JSON.parse` of the arguments the model chose, its own portal
+ * injection with no `get_data` guard, and a raw `error.message` handed back to
+ * the model. Wave #345 consolidated the application's three loops onto
+ * `src/lib/model-loop/run-tool-loop.ts` and not one of those fixes reached
+ * here, because the census that found the class and the guard that measured it
+ * were both scoped to `src/` and to `.ts`. What remains in this file is what is
+ * genuinely its own: the models it compares, the queries it runs them against,
+ * its MCP transport, its tool schema, and how it writes results up for scoring.
  *
  * THIS HARNESS IS PINNED TO OPENROUTER, AND THE APP IS NOT. Since
  * civic-ai-tools-website#30 the app reaches a configurable endpoint through
  * `src/lib/model-client.ts`, reading `MODEL_API_KEY` (with `OPENROUTER_API_KEY`
  * as its prior-era name) and `MODEL_API_BASE_URL` / `MODEL_API_KIND`. This
- * script reads none of that: it is `.mjs` run as bare `node`, so it cannot
- * import that TypeScript module, and it hardcodes OpenRouter's base URL and
- * requires `OPENROUTER_API_KEY` specifically. That is why the variable name
- * below differs from the one an operator now sets — the statement is accurate
- * about this script, not stale about the app. Repointing the harness at the
- * endpoint layer is tracked in civic-ai-tools#155 and is deliberately not a
+ * script reads none of that: it hardcodes OpenRouter's base URL and requires
+ * `OPENROUTER_API_KEY` specifically. That is why the variable name below
+ * differs from the one an operator now sets — the statement is accurate about
+ * this script, not stale about the app. Repointing the harness at the endpoint
+ * layer is tracked in civic-ai-tools#155 and is deliberately not a
  * find-and-replace.
  *
- * Usage:
+ * THE REASON RECORDED HERE FOR THAT PINNING WAS FALSE, and it is worth naming
+ * rather than quietly deleting: it said this file "is `.mjs` run as bare
+ * `node`, so it cannot import that TypeScript module." Node strips types from
+ * an imported `.ts` module — by default from v22.18, behind
+ * `--experimental-strip-types` before that — and the `runToolLoop` import below
+ * is the standing counter-example. Nothing about the extension prevented the
+ * import. The OpenRouter pinning is unfinished work (civic-ai-tools#155), not a
+ * constraint of the runtime.
+ *
+ * Usage, on Node >= 22.18 where type stripping is on by default:
  *   OPENROUTER_API_KEY=<your-key> SOCRATA_MCP_URL=https://your-mcp-host \
  *     node scripts/eval-models.mjs
+ *
+ * On Node 22.0 – 22.17 the same run needs the flag, because of that import:
+ *   OPENROUTER_API_KEY=<your-key> SOCRATA_MCP_URL=https://your-mcp-host \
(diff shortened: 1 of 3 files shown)


npstorey:
**Seat GO — head-bound to `fe237552813278e89b6f623023c99f4a12f541fb` (`fe23755`), conditional on one ruling below.** A moved head voids this.

- `git diff a544501...fe23755 --numstat`: 3 files, +253 −289 — equal to the PR API. `scripts/eval-models.mjs`, its test, and `model-call-registry.test.ts`; `run-tool-loop.ts` has an empty three-dot diff — the core was consumed, not edited.
- Check rollup on the exact head: `build / test / lint / typecheck`, `container image build`, `standalone build + asset check`, `DCO`, `Vercel` — all success, enumerated on the commit.
- Both commits (`39f7166`, `fe23755`) signed; `Signed-off-by` then `Co-Authored-By: Claude Opus 5 (1M context)`, contiguous, in order.
- **Criterion 1 is now true at this head:** `git grep -P '(args|toolArgs)\.portal\s*=[^=]'` over all tracked files, non-test → exactly one site, `run-tool-loop.ts:740`. The script imports the core (`eval-models.mjs:75`) and drives `runToolLoop` (`:448`); its own loop, parse, raw-error path and divergent injection are gone.
- **Criterion 5, on a runner:** run `33196411536` on probe PR #368 → `build / test / lint / typecheck` failure; #368 CLOSED, never merged, its branch gone. The registry test's universe is derived from `git ls-files -z --full-name` (`model-call-registry.test.ts:235`) and refuses an empty listing (`:243`) — scope true by construction, and the header no longer claims more than the code does.

**The hard-abort ruling (the owner's; the seat recommends accepting the removal).** P3 deleted the harness's hard-abort tier and its `hardAbort` / `hardAborts` / `totalHardAborts` fields rather than carry fields nothing can set. The seat's reasoning: a permanently-false field in a report is the asserted-default class this program treats as a defect (#312's shape, one directory over); the core bounds every tool result before it re-enters the transcript, which the old harness never did; and restoring the tier means editing the signed publishing path for a dev-only script whose output is never served or signed. **If the owner accepts the removal, this GO stands as written. If the owner rules to restore it, the head moves and this GO is void** — the ORCH re-reports.

Sequencing: merge now, per the ORCH — P5 is live on disjoint files and #367 already showed a base move is harmless here. Prediction accepted onto the record: `main` 1129 today, P3 lands 1131, the final head after P5 is 1131 + P5's additions with no name absent from any set, read from the runner.

P6 inherits three doc rows P3 made stale and disclosed: CLAUDE.md's model-loop scope sentence, `# pass 1092` → 1131, lint `4 problems` → 3.

PR #370, opened 2026-08-28
fix: the generated notebook stops claiming what it did not do (#340, #341, #342, #324)

### LINKED ISSUE #340: A helper-signature TypeError in a generated cell is reported to the reader as a live-data outage
## Summary

A generated notebook cell can be invalid Python at the call level — an argument the helper has no parameter for — and the failure is now reported to the reader as a live-data outage. The source is never contacted.

Produced by two individually correct changes meeting: one kept "append every argument the caller did not enumerate" in `pyKwargs`, the other wrapped fetch calls in `try/except Exception` so a failing fetch degrades instead of aborting the notebook.

## Why it is reachable, not hypothetical

`src/lib/mcp/tools.ts:43-46` documents `query` on `get_data` as *"For type=query: optional full-text search within data."* The model is invited to pass it.

`src/lib/notebook-author/helpers/fetch_socrata.py:13-19` declares `fetch_socrata(portal, dataset_id, *, select, where, group, …)` — keyword-only, **no `**kwargs`**. There is no `query` parameter.

`pyKwargs` appends `query` to the generated call. Python raises `TypeError` at call time. `guardedFetch` catches it.

## What the reader sees

```
Step 1: live data could not be fetched (TypeError); continuing with an empty table.
```

Immediately above it, the markdown cell reads *"Original execution returned 5 rows × 3 columns."* Below it, an empty table.

So the notebook states a precise row and column count for a fetch that never ran, attributes the emptiness to the data source, and presents both as a reproducible analysis. `docs/design-principles.md` Principle 3 is exactly about this: **false precision**. A cell that reports a live-data outage for a code defect is worse than one that throws, because it is confidently wrong about the cause.

The same shape occurs as `NameError` when a hand-rendered argument is absent.

## Suggested direction

Distinguish the two failure classes rather than merging them. A `TypeError`/`NameError` at call time is an authoring defect and should say so; a network or HTTP failure is the live-data case the guard was written for. A generated cell should not be able to claim a row count for a call that did not execute — the count line and the guard's fallback text disagree, and only one of them can be true.

npstorey:
## Phase P5 — the notebook renderer (Wave N8, lane 2)

**Model:** Opus 5 (1M context). **Base:** `cecc959`. **Branch:** `sprint-363/p5-notebook-renderer`.
**Head:** `8ab83f5`. Four commits.

Closes #340, #341, #342, and the notebook-identity half of #324 (per ruling D5).

### Diff stat

```
 src/lib/mcp/tools.ts                               |   8 +-
 .../helpers/fetch-socrata-contract.test.ts         | 400 +++++++++++++++++++++
 src/lib/notebook-author/helpers/fetch_socrata.py   |  90 ++++-
 src/lib/notebook-author/notebook-claims.test.ts    | 245 +++++++++++++
 src/lib/notebook-author/prompt.ts                  |  45 ++-
 src/lib/notebook-author/synthesize.test.ts         |   6 +-
 src/lib/notebook-author/synthesize.ts              |   7 +-
 src/lib/notebook-author/tool-to-cell.test.ts       | 345 ++++++++++++++++++
 src/lib/notebook-author/tool-to-cell.ts            | 374 +++++++++++++++++--
 src/lib/notebook-author/validate.ts                |  39 +-
 src/lib/notebook.ts                                |   9 +-
 11 files changed, 1516 insertions(+), 52 deletions(-)
```

### Blast zone — nine of eleven files are the chartered zone; **two are itemised additions**

Chartered and touched: `tool-to-cell.ts` + `tool-to-cell.test.ts`, `prompt.ts`, `validate.ts`,
`helpers/fetch_socrata.py`, `notebook.ts` (D5 only), `mcp/tools.ts` (the `query` description only),
plus two new test files matching `src/**/*.test.ts`.

**Additions, file by file:**

1. **`src/lib/notebook-author/synthesize.ts`** (+7/-2). Criterion 5's first half is not reachable
   without it. `buildCell0Source` has exactly one production caller, and cell 0 is written **before**
   the step cells exist — so the claim "this notebook contains a complete, reproducible analysis" can
   only become conditional if its caller passes what it knows. The edit is one import and one
   argument.
2. **`src/lib/notebook-author/synthesize.test.ts`** (+6/-2, one assertion). `:66` asserted
   `/Civic AI Data Analysis/` — the exact string ruling D5 removes. The test name is unchanged; the
   assertion now pins the D5 conditional (`/^# Data Analysis$/m` under no declared identity).

Same shape as the zone correction G2 §2 made for P2: a zone derived from the issue rather than from
the call graph. Neither file is touched by P3 (`scripts/eval-models.mjs`, its test, the registry
test) — lane disjointness holds.

### Evidence

| Check | Result |
|---|---|
| `npm ci` | clean (run before every measurement below) |
| `npm test` | `# tests 1156 · # suites 10 · # pass 1156 · # fail 0` — same on the runner, job `98940870819` of run `33198255316`, read from the job-log API |
| `npm run build` | `✓ Compiled successfully`, Route (app) table emitted |
| `npm run lint` | `✖ 4 problems (0 errors, 4 warnings)` — the baseline four, unchanged |
| `npm run check:compose-env` | `RESULT: PASS — every variable this profile reads can reach the container.` |

**Nothing lost.** TAP name sets, base `cecc959` vs branch head, sorted and `comm`-diffed:
**0 names present at the base and absent at the head**; **27 added** (13 in `tool-to-cell.test.ts`,
6 in `fetch-socrata-contract.test.ts`, 8 in `notebook-claims.test.ts`). 1129 → **1156**.

### The ruling, implemented

**#340.** The helper gains `query` and implements the service's split itself — it talks to the
portal directly, and a reader can re-run the cell with their own arguments. On the SoQL branch the
cell renders `query=` alone; the superseded clauses are omitted, not written as arguments with no
effect. The generated comment, verbatim, for the fixture the tests drive:

```python
# `query` here is a full SoQL statement, so the data source applied it as the
# whole query: select / where / group / order — and limit / offset — are not
# sent alongside it. They are omitted below rather than written as arguments
# that would have no effect on the rows this cell returns.
# Superseded on this call, for that reason: select, where, limit.
# The row count is therefore bounded by the statement's own LIMIT, or by the
# portal's default page size when the statement carries none.
# One deliberate difference from the original run: this helper always requires
# an explicit dataset_id and never derives one from `query`.
```

Rider (a) — the cross-repo coupling is pinned by name in both test files: the service's regex
transcribed literally, `socrata-tools.ts:546` at `116f46ce1e84e3608014599f9b63ea01acfd913a`, the
fixture from `search.test.ts:165`, and a header stating that a service-side change is an issue on
**this** repository rather than a silent divergence. Rider (b) — the row bound is named, with its own
red demonstration below. Rider (c) — the helper raises `SocrataDatasetIdRequired` and never infers;
the divergence from `:531` is stated in the same comment block; a query call that named no dataset
renders a note instead of a cell that always raises, and the `unknown` placeholder is gone.

Beyond the letter of #340, the same bound now applies to **every** argument rather than to `query`
alone: a key no helper parameter matches is disclosed in the cell instead of emitted, and a test
reads the helpers' real signatures (and asserts they carry no `**kwargs`) so neither list can drift.

**#341.** The cover claim is conditional on at least one fetch surviving, and
`validateExecutedNotebook` gains a third validator that reports a notebook in which no step re-runs
a fetch. The two are computed by different means on purpose — the cover cell from the tool calls
(it is written first), the validator from the cells (so it cannot be satisfied by a claim) — and a
test pins that they still answer the same question.

**#342.** `describeAttempt` dispatches on name **and** `args.type`. Every `unknown` placeholder in
that function is gone, in all four tool branches.

**#324 (D5).** Both hardcoded titles become the `#258 A2` conditional: the declared platform title,
or the description alone.

### Red demonstrations — eleven, each on the test that names it

| # | Reverted | Red |
|---|---|---|
| R1 | superseded kwargs back into the cell | `#340: a SoQL query renders query= and no superseded kwarg` — `select= had no effect on this call and must not appear in the cell` |
| R2 | the row-bound comment lines deleted | `#340: the SoQL comment names the supersession AND the row bound` |
| R3 | `SOQL_QUERY_SNIFF` → `/^select/i` | `#340: our sniff IS the service's sniff, character for character` |
| R4 | the Python sniff → `query.upper().startswith("SELECT")` | `#340: the helper applies the sniff itself…` — `wrong branch for "   SELECT count(*) AS n"` |
| R5 | the cover claim unconditional again | `#341: a notebook whose every fetch failed…`, `#341: the cover claim and the validator answer the same question` |
| R5b | the same, re-run after the copy fix | the two above **plus** `#341: a notebook that never fetched is not told its requests failed` |
| R6 | the third validator dropped | `#341: validateExecutedNotebook reports an all-failed notebook` |
| R7 | `query` removed from the helper (the base state) | all six contract tests — `TypeError: fetch_socrata() got an unexpected keyword argument 'query'` |
| R8 | the helper infers `dataset_id` from `query`, as the service does | `#340 (rider c): dataset_id is required and never inferred` |
| R9 | `describeAttempt` back to name-only dispatch | both `#342` tests |
| R10 | the fetch detector matches the helper NAME | `#341: the detector is not fooled by the helper-definitions cell` |

R4 and R10 are the two that matter most: both are re-spellings that pass a naive check. R4 diverges
only on leading whitespace; R10 counts the helper-**definitions** cell and would report every
all-failed notebook as reproducing a fetch — the check passing on exactly the document it exists to
reject.

No test instrument in this phase uses timers, `.unref()`, or any asynchronous guard, so no red case
can be converted into a cancelled file (the shape P2 hit).

### One defect this phase introduced and then removed (commit `8ab83f5`)

The first version of the no-fetch cover text said *"every request to a live data source returned no
data"* — true of the all-failed notebook it was written for, false of the other document that
reaches the same branch: an analysis answered from discovery alone, whose catalog searches all
**succeeded** and which simply has no fetch step. That is this phase's own defect class one step to
the left. The text now says what is true of the document — no step re-runs a request — and a test
drives the discovery-only notebook. Found by re-reading the diff; the test came second.

### Flagged, not fixed

1. **`src/lib/notebook.ts:60-72` carries the #340 shape** — `buildSocrataUrl` builds a direct
   `/resource/<id>.json?$select=…` URL and ignores `query` and `offset` entirely. My contract limits
   that file to D5. P7's cold read is already pointed at it.
2. **`renderOpenContextQueryCell` silently drops every arg but `resource_id`/`filters`/`limit`.**
   Satisfies "does not pass it", but with no disclosure — the CKAN half of the class #340 names.
3. **`fetch_socrata`'s default `limit=1000` vs the service's `10`** for `type=query`. Pre-existing;
   a cell carrying no explicit `limit` already reproduces a wider page than the original run.
   Changing it would move row counts in existing notebooks, so it is a decision, not a fix.
4. **`get_data`'s schema advertises no `having` or `q`, and does not set `additionalProperties`.**
   The model can therefore still send a key nothing expects — now disclosed rather than fatal.
5. **`docs/` is untouched**: the helper's `query` parameter and the `dataset_id` divergence are
   described nowhere outside the code. A P6-adjacent docs line.
6. **A partial-failure notebook still carries the full claim.** The conditional is binary, as the
   contract specifies; a notebook where three of four fetches failed says "complete, reproducible
   analysis" while its per-step notes say otherwise. Worth a ruling, not a silent widening.

### Runner

All six checks pass on head `8ab83f5`: `build / test / lint / typecheck`, `container image build`, `Vercel`, `DCO` (the four required) plus `standalone build + asset check` and `Vercel Preview Comments`. The standalone job is the one that re-copies the edited `.py` helper and byte-compares it, so the helper change is covered on a host as well as in the container.

### DIFF
--- a/src/lib/mcp/tools.ts
+++ b/src/lib/mcp/tools.ts
@@ -42,7 +42,13 @@ Examples:
           },
           query: {
             type: 'string',
-            description: 'For type=catalog: search query. For type=metadata: the dataset ID. For type=query: optional full-text search within data.',
+            // #340: this used to describe only the $q branch. The data-access
+            // handler splits on whether the value starts with SELECT, and the
+            // SoQL branch supersedes the individual clauses AND drops
+            // limit/offset — so a model reading the old text could send both a
+            // SELECT and a limit and be silently given neither bound it asked
+            // for. Both branches are stated here because both are reachable.
+            description: 'For type=catalog: search query. For type=metadata: the dataset ID. For type=query: either a full SoQL statement starting with SELECT, which is applied as the entire query and supersedes select/where/order/group (limit and offset are not applied either — bound the rows with the statement\'s own LIMIT), or a search phrase, applied as a full-text search within the data alongside the other clauses.',
           },
           dataset_id: {
             type: 'string',
--- a/src/lib/notebook-author/helpers/fetch-socrata-contract.test.ts
+++ b/src/lib/notebook-author/helpers/fetch-socrata-contract.test.ts
@@ -1,0 +1,400 @@
+// The contract between a generated cell and the helper it calls (Wave N8 P5,
+// #340).
+//
+// Two halves, and the defect lived in the gap between them:
+//
+//   1. `src/lib/mcp/tools.ts` advertises `query` on `get_data`, the model
+//      sends it, and `tool-to-cell.ts` renders it into a `fetch_socrata(...)`
+//      call. At the base `fetch_socrata` had no `query` parameter, so the cell
+//      raised `TypeError` on execution and the reader was told live data could
+//      not be fetched — in a notebook whose cover text called itself
+//      reproducible.
+//   2. The helper is not a wrapper around the data-access service: it talks to
+//      the portal directly. So it cannot inherit the service's precedence for
+//      `query`; it has to implement the same rule itself. A reader can re-run
+//      this cell with different arguments, and the mapping has to hold for
+//      THEIR arguments too, not only for the ones we rendered.
+//
+// THE COUPLING THIS FILE PINS, AND WHAT TO DO WHEN IT MOVES.
+// `_is_full_soql_query` in fetch_socrata.py is a copy of a regular expression
+// that lives in ANOTHER repository:
+//
+//     socrata-mcp-server/src/tools/socrata-tools.ts:546
+//     at commit 116f46ce1e84e3608014599f9b63ea01acfd913a
+//
+//         if (queryField && /^\s*select/i.test(queryField)) { … }
+//
+//   matching  → `$query` alone; select/where/order/group/having/q are set
+//               aside (:547-553) and the request carries neither $limit nor
+//               $offset (:283-293);
+//   otherwise → `$q`, with every other clause preserved (:555-557);
+//   and, third, a non-SoQL `query` with no dataset_id BECOMES the dataset id
+//               (:531) — the one behaviour this helper deliberately does not
+//               reproduce.
+//
+// IF THE SERVICE CHANGES ITS SNIFF, THE FIX IS AN ISSUE ON THIS REPOSITORY, so
+// that this copy, the TypeScript predicate in `tool-to-cell.ts` and the
+// service all move together. A change made on one side alone is a silent
+// divergence, and it surfaces only in a published notebook — where a reader is
+// told a `limit=` applied that never did.
+//
+// Run with: npm test
+
+import { test } from 'node:test';
+import assert from 'node:assert/strict';
+import { spawnSync } from 'node:child_process';
+import fs from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { HELPER_PARAMETERS, renderFetchToolCell } from '../tool-to-cell.ts';
+import { mcpTools } from '../../mcp/tools.ts';
+
+const HELPERS_DIR = path.dirname(fileURLToPath(import.meta.url));
+
+function helperSource(file: string): string {
+  return fs.readFileSync(path.join(HELPERS_DIR, file), 'utf8');
+}
+
+/**
+ * Stub `requests` and `pandas` into `sys.modules`, then execute the helper
+ * source as written. Stdlib only — the suite must not need the scientific
+ * stack installed to check what the helper sends.
+ *
+ * The captured request deliberately excludes headers: they carry
+ * `X-App-Token` when the environment supplies one, and a test that printed
+ * them would print a credential into a CI log.
+ */
+const PY_PREAMBLE = [
+  'import json, sys, types',
+  'captured = {}',
+  'class _Response:',
+  '    def raise_for_status(self): return None',
+  '    def json(self): return []',
+  'def _get(url, params=None, headers=None, timeout=None):',
+  "    captured['url'] = url",
+  "    captured['params'] = params",
+  '    return _Response()',
+  "requests_stub = types.ModuleType('requests')",
+  'requests_stub.get = _get',
+  "pandas_stub = types.ModuleType('pandas')",
+  'pandas_stub.DataFrame = lambda data=None: data',
+  "sys.modules['requests'] = requests_stub",
+  "sys.modules['pandas'] = pandas_stub",
+  'ns = {}',
+  "exec(compile(sys.stdin.read(), 'helper.py', 'exec'), ns)",
+].join('\n');
+
+/** Call `fetch_socrata(**kwargs)` and report the query string it would send. */
+const PY_CALL = [
+  PY_PREAMBLE,
+  'kwargs = json.loads(sys.argv[1])',
+  'try:',
+  "    ns['fetch_socrata'](**kwargs)",
+  "    print(json.dumps({'raised': None, 'url': captured.get('url'), 'params': captured.get('params')}))",
+  'except Exception as err:',
+  "    print(json.dumps({'raised': type(err).__name__, 'message': str(err)}))",
+].join('\n');
+
+/**
+ * Report a helper's parameter names, plus whether it declares `*args` /
+ * `**kwargs`, by parsing the .py rather than importing it.
+ *
+ * A structural read of the source, not `inspect.signature` on an imported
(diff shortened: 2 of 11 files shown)


npstorey:
**Seat GO — head-bound to `8ab83f5` (PR #370).** A moved head voids this.

- `git diff ffaffb7...8ab83f5 --numstat`: 11 files, +1516 −52 — equal to the PR API. Lane boundary clean: `scripts/eval-models.mjs` and `src/lib/model-loop/` have an empty three-dot diff on this branch. The two zone additions (`synthesize.ts` and its test — the cover cell is built before the step cells exist, so its one caller must pass what it knows) are ratified: a call-graph zone, itemised at the gate.
- Check rollup on the exact head: `build / test / lint / typecheck`, `container image build`, `standalone build + asset check`, `DCO`, `Vercel` — all success, enumerated on the commit.
- Four commits, all signed; each carries `Signed-off-by` then `Co-Authored-By`, contiguous, in that order.
- **The cross-repo pin (rider a), read in the files:** `tool-to-cell.ts:192-193`, `fetch-socrata-contract.test.ts:22-23` and `tool-to-cell.test.ts:508-509` name `socrata-mcp-server/src/tools/socrata-tools.ts:546` at commit `116f46ce1e84e3608014599f9b63ea01acfd913a`. The seat checked the server: that commit is on the server's `origin/main` (two behind its tip), and its line 546 is `if (queryField && /^\s*select/i.test(queryField))` — the pin is real, fetchable, and points at the sniff it mirrors.
- P3's merge `ffaffb7` has second parent `fe23755`, the SHA that GO named; the late `rollback/pre-n8-p3` and `-p5` tags exist on origin at the base both phases branched from.

**Two rulings the seat recommends, for the owner:** (2) file the partial-failure cover text as its own issue rather than widening P5 — criterion 10 said "every fetch failed" and P5 met it as written; where "complete" ends for a three-of-four notebook is a product judgement about a published artifact, and a phase does not make it. (3) P4b ∥ P6 from the post-P5 head as the last pair, P6 branching only after P5 lands because D9's add-the-schemas branch edits `src/lib/mcp/tools.ts`, which P5 touched. Disjoint by file thereafter; the parallel-pair prediction rule applies to the second merge.

Prediction accepted onto the record: merged head 1158.

PR #372, opened 2026-08-28
fix: absent token usage is absent from every trace span (#312)

### LINKED ISSUE #312: Trace spans record 0 for absent token usage — a false claim inside the signature

npstorey:
Wave N8, Phase P4b (lane 1). Base `f117665` (tag `rollback/pre-n8-p4b`). Closes #312.

## Blast zone

Two files, both inside the declared zone. Nothing else in the repository was created, edited or deleted.

```
 src/lib/model-loop/absent-usage.test.ts | 516 ++++++++++++++++++++++++++++++++
 src/lib/model-loop/run-tool-loop.ts     | 113 ++++++-
 2 files changed, 613 insertions(+), 16 deletions(-)
```

`src/lib/model-loop/absent-usage.test.ts` is new and matches `src/**/*.test.ts`, which the zone admits. Nothing P6 owns was touched — in particular `model-call-registry.test.ts`, `socrata-skill.ts`, `preamble-advertised-tools.test.ts`, `mcp/tools.ts`, `evidence/[slug]/evaluate/route.ts`, `evidence/adversarial-eval.ts` and the docs are all unchanged. `test-harness.ts` is unchanged (see the flag below).

## What changed

Six sites, two mechanisms, one construction site now.

**Mechanism one** — both `llm_inference` sites — coalesced absence to zero with an inline `|| 0` on `response.usage`.

**Mechanism two** — the `synthesis` span, the answering turn that produces the published answer — reached the same false zero by a different route, which is why a grep for the `|| 0` shape could not see it: `finalPromptTokens` / `finalCompletionTokens` were `0`-initialised numbers that simply stayed `0` when no usage arrived. They now start `undefined`.

Both now go through one private `responseTokenAttributes(usage)`, which spreads a count in **only when a count was reported**, on each of the two counts independently.

**Omission means the key is never built, not that its value is `undefined`.** Measured against this repo's own dependencies: `TraceBuilder` maps an attribute record through `Object.entries` into an array of `{key, value}` pairs, so a key whose value is `undefined` does **not** vanish the way an object property would — it survives canonicalization as `{"key":"gen_ai.response.prompt_tokens","value":{}}`. A valueless attribute under the signature is worse than the zero it replaced. This inverts the trap #312 names, and the canonical-bytes assertion catches it either way.

**A genuine zero is still a measurement.** The streaming path's truthy guards (`if (chunk.usage.prompt_tokens)`) would have erased an endpoint that truthfully reported `0`; they become absence checks bounded by `Number.isFinite`, which keeps the old guards' one useful effect — a `NaN` never lands in a signed trace — while admitting the zero.

Newly captured traces only. No re-emission, no lifecycle events, no migration.

## The census, re-derived on the branch

`git grep -n -P "gen_ai\.response\.(prompt|completion)_tokens"` over all tracked files, no pathspec:

```
src/lib/model-loop/absent-usage.test.ts:6:// `gen_ai.response.prompt_tokens: 0` for a call the endpoint said nothing
src/lib/model-loop/absent-usage.test.ts:40://     `{"key":"gen_ai.response.prompt_tokens","value":{}}`, a valueless
src/lib/model-loop/absent-usage.test.ts:76:const PROMPT_TOKENS = 'gen_ai.response.prompt_tokens';
src/lib/model-loop/absent-usage.test.ts:77:const COMPLETION_TOKENS = 'gen_ai.response.completion_tokens';
src/lib/model-loop/run-tool-loop.ts:431: * `{"key":"gen_ai.response.prompt_tokens","value":{}}`, measured, and the JCS
src/lib/model-loop/run-tool-loop.ts:459:    ...(promptTokens !== undefined ? { 'gen_ai.response.prompt_tokens': promptTokens } : {}),
src/lib/model-loop/run-tool-loop.ts:460:    ...(completionTokens !== undefined ? { 'gen_ai.response.completion_tokens': completionTokens } : {}),
```

The only emission sites are `:459-460`; every other hit is prose or a test constant. Six unconditional emission sites became one conditional pair.

## Red demonstrations

Criteria 2 and 3 pull in opposite directions, so both were shown red — 2 against the base, 3 against a deliberately-wrong fix, because the base already passes it.

**Criterion 2 — the synthesis span, RED at `f117665`** (the new suite run before the source change):

```
not ok 3 - #312: an answering turn whose endpoint reports no usage leaves the synthesis span with neither count (blocking)
  error: the answering turn reported no usage: span "synthesis" carries gen_ai.response.prompt_tokens = {"intValue":"0"}
not ok 4 - #312: an answering turn whose endpoint reports no usage leaves the synthesis span with neither count (stream)
  error: the answering turn reported no usage: span "synthesis" carries gen_ai.response.prompt_tokens = {"intValue":"0"}
```

That case has the **tool turns reporting usage**, so a fix reaching only mechanism one leaves the `llm_inference` spans correct and this case still red. Ten cases, 3 pass / 7 fail at the base.

**Criterion 3 — a genuine zero, RED under a falsy-check fix.** `reportedCount` temporarily replaced by `return value ? Number(value) : undefined`:

```
ok 1 - ... an endpoint reporting no usage leaves no token attribute on any span (blocking answering turn)
ok 2 - ... (stream answering turn)
ok 3 - ... the synthesis span with neither count (blocking)
ok 4 - ... (stream)
not ok 5 - #312: an endpoint reporting a genuine zero still gets 0 on every span (blocking answering turn)
not ok 6 - #312: an endpoint reporting a genuine zero still gets 0 on every span (stream answering turn)
# tests 10 / # pass 8 / # fail 2
```

**Criterion 4 — omission by `undefined`, RED.** The conditional spreads temporarily replaced by unconditional `'gen_ai.response.prompt_tokens': promptTokens as number`:

```
not ok 1 - ... error: the endpoint reported no usage: span "llm_inference" carries gen_ai.response.prompt_tokens = {}
not ok 3 - ... error: the answering turn reported no usage: span "synthesis" carries gen_ai.response.prompt_tokens = {}
# tests 10 / # pass 4 / # fail 6
```

Both counterfactuals were reverted; `git grep -c COUNTERFACTUAL` over the branch returns nothing.

## The returned `usage` accumulators — argued, not silently left

**Decision: unchanged, and pinned by a test.**

- It is a **sum over turns**, and "absent contributes nothing to a total" is the correct semantic for a sum. A per-span count is a single call's measurement, where absent and zero are exactly distinguishable; a run-wide total where one turn reported and another did not has no honest scalar other than "the sum of what was reported."
- **Both consumers already omit at their own boundary**, which is where absence is expressible: `packager.ts:476` (`promptTokens: input.tokenUsage.promptTokens || undefined`) for the signed `cost` block, and `openrouter-streaming.ts:297-298` for the wire. So a run with no usage anywhere already produces a package with the counts absent.
- Widening `ToolLoopResult['usage']` to `number | undefined` would change the return type for every caller — `openrouter-streaming.ts`, `compare-loop.ts`, `replay-loop.ts`, `evidence/[slug]/replay/route.ts` and `AttestationDialog.tsx` (whose `tokenUsage` requires numbers) — all outside this phase's zone, to express something no consumer currently needs.

The test `#312: with no usage reported anywhere the returned totals are zero, and the counts are omitted at the boundary` states this on the record, so a later phase that changes it has to change that line and say why.

## Flagged, not fixed

1. **`src/lib/model-loop/test-harness.ts` cannot express absent usage.** `startScriptedModelServer` hardcodes `{prompt_tokens: 10, completion_tokens: 5}` on every reply. The right fix is a `usage?: Record<string, number> | null` field on `ScriptedReply`, after which `startUsageScriptedServer` in the new test file deletes entirely. Out of zone for P4b, so the capability is local and the duplication is flagged rather than taken.
2. **`src/lib/openrouter-streaming.ts:209-212`** — the tool-free streaming path tracks `if (chunk.usage?.total_tokens)` and reports `tokens_used: tokensUsed` with `tokensUsed` initialised to `0`. Same class as #312 (absent usage reported as zero) at the wire/display layer rather than on a span, so not under a signature; not fixed, out of zone.
3. **`packager.ts:476` and `openrouter-streaming.ts:297-298` use a falsy check** (`|| undefined`) rather than an absence check. Correct in practice at the sum layer — a run whose endpoints all reported would never total zero prompt tokens — but it is the same shape criterion 3 exists to prevent, one layer up. Out of zone.
4. **The token budget is unenforceable when usage is absent.** `cumulativeTokens` stays `0`, so `maxCumulativeTokens` never trips on an endpoint that reports nothing. Pre-existing, arguably correct (a budget cannot be enforced on an unmeasured quantity), unexamined by any test. Not this phase.
5. **`CLAUDE.md`'s command table is stale at `f117665`**: it documents `npm test` as `# pass 1092` (actual base: 1158) and `npm run lint` as `✖ 4 problems (0 errors, 4 warnings)` (actual base, verified by stashing this branch's changes: `✖ 3 problems (0 errors, 3 warnings)`). Docs are out of zone.

## Contract premises — all six line references verified, one contract-layer detail corrected

Every `file:line` in the phase contract survived the check, at `f117665`:

- `run-tool-loop.ts:637-638`, `:842-843`, `:965-966` — exact, all six attributes.
- `:916-917` locals initialised to `0`; `:938` `if (chunk.usage)`; `:940-941` the assignments; `:953-954` the non-stream `|| 0`; `:645` `cumulativePromptTokens` init; `:849` the add; `:958-959` the final add — all exact.
- `packager.ts:470-478` carries the "Zero is not a measurement (website#30 P3, §2.5)… Absent usage is now absent" comment, the comment opening at `:470` and the two count lines at `:476-477`.
- Baseline `# tests 1158 / # suites 10 / # pass 1158 / # fail 0` reproduced locally at `f117665`, identical to run `33199189643`.

**Corrected:** #312's fix-shape note says an `undefined` property is dropped by the canonicalizer so a `hasOwnProperty` assertion "fails while the behaviour is correct." For a plain object that is true. For a **trace span** it is the opposite, and it was measured here: attributes are an array of `{key, value}` records built by `Object.entries`, so an `undefined` value leaves the key in place with an empty value object, and the canonicalizer keeps it. The instruction that follows — assert over canonical bytes — is right for both, and is what the suite does.

## Checks

`npm ci` first. All five green.

**`npm test`**

```
1..1135
# tests 1168
# suites 10
# pass 1168
# fail 0
# cancelled 0
# skipped 0
# todo 0
# duration_ms 1995.54175
```

Base was `# tests 1158 / # suites 10 / # pass 1158 / # fail 0`; +10, exactly the ten cases added. Test names were captured before and after and diffed: **zero names removed**, ten added.

**`npm run build`** — exit 0.

```
▲ Next.js 16.3.0 (Turbopack)
✓ Running next.config.ts took 10ms

✓ Compiled successfully in 249ms
  Running TypeScript ...
  Finished TypeScript in 984ms ...
✓ Generating static pages using 9 workers (32/32) in 186ms
```

followed by the Route (app) table. The two Turbopack warnings are pre-existing (`src/lib/notebook-author/helpers/index.ts:26`, dynamic filesystem access) and untouched by this change.

**`npm run typecheck`** — no output, exit 0.

**`npm run lint`** — exit 0.

```
✖ 3 problems (0 errors, 3 warnings)
```

Three warnings, all pre-existing (`AttestationDialog.tsx:8`, `RenderingCellOutputs.tsx:96`, `bpmn/animation.ts:8`); the base measures 3 as well.

**`npm run check:compose-env`**

```
```

## Provenance

One commit, both trailers contiguous and in order, verified with `git log --format='%(trailers:only)'`:

```
Signed-off-by: Nathan Storey <npstorey@users.noreply.github.com>
Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>
```

Exactly one `Signed-off-by`, matching the author `Nathan Storey <npstorey@users.noreply.github.com>` exactly. The message body was written without trailers and composed by `-s` plus `--trailer`.

Model: **Opus 5** (`claude-opus-5[1m]`).

## On the runner

Run `33200662901`, at `70d19dc`. All five required checks pass.

```
DCO                            pass
Vercel                         pass
Vercel Preview Comments        pass
build / test / lint / typecheck  pass  1m30s
container image build          pass  59s
```

`build / test / lint / typecheck` on Node `v22.23.2` reproduces the local figure exactly:

```
# tests 1168
# suites 10
# pass 1168
# fail 0
# cancelled 0
# skipped 0
# todo 0
```

### DIFF
--- a/src/lib/model-loop/absent-usage.test.ts
+++ b/src/lib/model-loop/absent-usage.test.ts
@@ -1,0 +1,516 @@
+// Acceptance tests for #312: absent token usage is ABSENT from the trace — on
+// every span — and a genuine zero still is not.
+//
+// WHY THIS IS A PROPERTY OVER SPANS AND NOT A FIX TO FOUR LINES. The trace
+// goes inside the signed record package, so a span asserting
+// `gen_ai.response.prompt_tokens: 0` for a call the endpoint said nothing
+// about is a false statement under a signature. Wave N4 P3 fixed exactly that
+// one layer up for `cost.promptTokens` (`packager.ts`: "Zero is not a
+// measurement… Absent usage is now absent"), leaving a package that could
+// carry `cost` with the counts correctly absent while the span beside it
+// claimed zero.
+//
+// The issue named four sites. There are SIX, across TWO mechanisms, and the
+// second is why a grep for the `|| 0` shape could not find it:
+//
+//   mechanism one — `llm_inference`, both sites: an inline `|| 0` on
+//     `response.usage`, coalescing absence to zero at the span;
+//   mechanism two — `synthesis`, the answering turn that produces the
+//     PUBLISHED answer: two locals initialised to `0` that simply stay `0`
+//     when no usage arrives (the streamed path guarded `if (chunk.usage)`,
+//     the blocking path used its own `|| 0`). `0` from "never set" and `0`
+//     from "the endpoint reported zero" are the same bytes unless something
+//     tracks the difference.
+//
+// So the cases below drive the REAL loop to completion and assert over EVERY
+// span the trace finalizes, not over the two lines an issue pointed at. The
+// synthesis span gets its own case, with the tool turns REPORTING usage, so
+// that a fix addressing only mechanism one goes red here rather than passing.
+//
+// AND ASSERTED OVER THE CANONICAL FORM, which #312 asks for explicitly. Two
+// different traps make a naive presence check the wrong instrument:
+//
+//   - `produce-core` leaves an `undefined` property present on an object and
+//     the JCS canonicalizer drops it, so a `hasOwnProperty` assertion can fail
+//     while the behaviour is correct;
+//   - and in the OTHER direction, measured here: `TraceBuilder` turns an
+//     attribute record into an ARRAY of `{key, value}` pairs via
+//     `Object.entries`, so setting an attribute to `undefined` does NOT remove
+//     it — it survives canonicalization as
+//     `{"key":"gen_ai.response.prompt_tokens","value":{}}`, a valueless
+//     attribute inside the signature, which is worse than the zero it
+//     replaced.
+//
+// Every case therefore asserts over `jcs(trace)` — the bytes the signature
+// actually covers — as well as over the span structure.
+//
+// The RED baseline, measured at f117665 before the fix, driving the
+// tool-then-silence script with an endpoint that reports no usage at all:
+// every `llm_inference` span carried `prompt_tokens` `intValue: "0"` and the
+// `synthesis` span carried it too, by the second mechanism. And with a
+// falsy-check fix in place of the absence check, the genuine-zero case below
+// goes red instead — the same defect pointing the other way.
+//
+// No live endpoint, no credential, no MCP server, no database. Every key value
+// is an obviously fake fixture and every address is loopback.
+//
+// Run with: npm test
+//   (or: node --test --experimental-strip-types src/lib/model-loop/absent-usage.test.ts)
+
+import { test, beforeEach, afterEach } from 'node:test';
+import assert from 'node:assert/strict';
+import { createServer, type Server } from 'node:http';
+import type { AddressInfo } from 'node:net';
+import { runToolLoop, type ToolLoopOptions } from './run-tool-loop.ts';
+import { _resetDefaultModelClientForTests, getModelClient } from '../model-client.ts';
+import { TraceBuilder, CIVICAITOOLS_TRACE_CONFIG } from '../evidence/trace.ts';
+import { jcs } from '../evidence/canonicalization.ts';
+
+const FIXTURE_KEY = 'not-a-real-key-p4b-absent-usage-fixture';
+const PROMPT = 'How long do these requests take to close?';
+const REAL_ANSWER =
+  'Across both years the portal recorded 4,812 requests of this type. The median time to close ' +
+  'was 9 days, and 71% closed within 14 days (dataset abcd-1234).';
+const ONE_ROW = JSON.stringify({ data: [{ request_type: 'A', days_to_close: 9 }], total_rows: 1 });
+
+const PROMPT_TOKENS = 'gen_ai.response.prompt_tokens';
+const COMPLETION_TOKENS = 'gen_ai.response.completion_tokens';
+
+const savedEnv: Record<string, string | undefined> = {};
+const ENV_KEYS = ['OPENROUTER_API_KEY', 'OPENAI_API_KEY', 'MODEL_API_BASE_URL'];
+
+beforeEach(() => {
+  for (const k of ENV_KEYS) {
+    savedEnv[k] = process.env[k];
+    delete process.env[k];
+  }
+  _resetDefaultModelClientForTests();
+});
+
+afterEach(() => {
+  for (const k of ENV_KEYS) {
+    if (savedEnv[k] === undefined) delete process.env[k];
+    else process.env[k] = savedEnv[k];
+  }
+  _resetDefaultModelClientForTests();
+});
+
+// --- The instrument --------------------------------------------------------
+
+/**
+ * A scripted endpoint whose `usage` reporting is the variable.
+ *
+ * `test-harness.ts`'s `startScriptedModelServer` is this family's instrument
+ * everywhere else and would be the right home for this, as a `usage` field on
+ * `ScriptedReply` — it hardcodes `{prompt_tokens: 10, completion_tokens: 5}`
+ * on every reply and offers no way to express "the endpoint reported none",
+ * which is the entire subject of this file. That file is outside this phase's
+ * blast zone, so the capability lives here and the duplication is flagged
+ * rather than taken: see the phase report.
+ *
+ * Real HTTP and real SSE on purpose. The claim is that the loop records
+ * nothing when the ENDPOINT reports nothing, and a hand-built async iterable
+ * of chunk objects would assert that about a fixture rather than about the
+ * client's own parsing of a response that genuinely omits the field.
+ */
+interface UsageScript {
+  content?: string | null;
+  toolCalls?: { id: string; name: string; args: Record<string, unknown> }[];
+  /**
+   * The `usage` object the endpoint returns for this reply. Omitted here means
+   * omitted on the wire: no `usage` key in the JSON body, and no usage frame
+   * in the SSE stream at all — which is what an endpoint that reports nothing
+   * actually sends.
+   */
+  usage?: Record<string, number>;
+}
+
+function startUsageScriptedServer(
+  replies: UsageScript[],
+): Promise<{ server: Server; url: string; requests: number }> {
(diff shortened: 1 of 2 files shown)


npstorey:
Seat GO - head-bound to 70d19dc (PR #372). A moved head voids this. Verified by the seat on 2026-08-28 from the remote ref and the runner: three-dot diff +613/-16 over two files, equal to the PR API; P6 files (src/lib/mcp, the evidence routes, docs, CLAUDE.md) have an empty three-dot diff on this branch; all four required checks plus Vercel green on this exact commit; one signed commit with Signed-off-by then Co-Authored-By, contiguous. The structural claim read in the blob: one conditional pair at run-tool-loop.ts:459-460 keyed on !== undefined, and zero remaining || 0 on any token attribute. The #312 fix-shape correction (a valueless attribute under a signature is worse than the zero it replaced; assert over canonical bytes) is accepted and is the second time this wave a property-shaped criterion caught what an issue prescribed. Merge now; P6 continues on disjoint files. File the openrouter-streaming.ts:209-212 tokens_used:0 finding as its own issue - same class at the wire layer, outside every N8 zone.

PR #373, opened 2026-08-28
fix: the two tools the prompt described, one rubric call, and five doc claims (#323, #348, #354)

### LINKED ISSUE #323: System prompt advertises two Socrata tools that are never added to the tool list

npstorey:
Wave N8 phase **P6** (lane 1), branched from `f117665` (`rollback/pre-n8-p6`). Closes #323, #348, #354.

**Model: Opus 5.** Diff: 8 files, +508 / −120.

```
 CLAUDE.md                                      |  33 ++-
 docs/instance-setup.md                         |  35 ++-
 docs/project-plan.md                           |  19 ++
 src/app/api/evidence/[slug]/evaluate/route.ts  |  48 ++--
 src/lib/mcp/preamble-advertised-tools.test.ts  |  75 -------
 src/lib/mcp/prompt-advertised-tools.test.ts    | 300 +++++++++++++++++++++++++
 src/lib/mcp/tools.ts                           |  61 +++++
 src/lib/model-loop/model-call-registry.test.ts |  57 ++++-
```

**Blast zone honoured.** Every file above is in the declared zone. `preamble-advertised-tools.test.ts` → `prompt-advertised-tools.test.ts` is a `git mv` of a zone file, under the zone's "new test files as needed, matching `src/**/*.test.ts`" clause — a guard whose name states a narrower reach than it has is this wave's own failure one level down. **`src/lib/model-loop/run-tool-loop.ts` and `run-tool-loop.test.ts` were read but never written** (P4b owns them). Nothing in `socrata-mcp-server` or `civic-ai-tools` was modified — `git status --porcelain` in each returns zero lines. No `.env*` file was read.

---

## 1. #323 — no surface reaching the model names a tool it cannot call

**Ruling D9 re-verified in the server's source before acting on it.** `search` and `fetch` are implemented, so the branch is add-the-schemas. Two line citations in the contract were off by one and are corrected below.

`mcpTools` gains both, mirroring the server's schemas exactly: one required property each (`query` / `id`) and `additionalProperties: false`. Neither accepts a portal or a domain, which is also why the loop core's portal injection stays scoped to `get_data`.

**The guard became the property.** Its corpus is now every constant `composeSkillPrompt` can put in front of the model — the preamble, the fallback, `getSkillForPortal`'s output for every portal it knows, **and `DATA_COMMONS_SKILL` and `BOSTON_OPENCONTEXT_SKILL`**, which reach the same prompt through the same registry and which the contract's "all three" did not name. Two directions, because they fail differently: an *invented* name is caught by extracting from four anchored positions (a `Tools:` list, a bold lowercase identifier, a lowercase-identifier table cell, a snake_case word before "tool"); a *known-but-uncallable* name is caught structurally — every routable tool name must have a schema, and no constant may contain, as a word, a routable name that does not. That second check is what #323 actually was, and no extractor was going to catch "start with search to discover datasets".

Blind spots are named in the header rather than left to be inferred, including one that costs the guard real reach.

## 2. #348 — one rubric call

`runAdversarialEval` already took `opts.apiKey`, so the route needed no signature change — only to stop assembling its own copy. `ALLOWED_MODEL_CALLERS` drops to five in the same commit. The route's structural check (`parsed.ok` → 502) is preserved, and the comment now says why it is there: it is what makes the bare `content || ''` extraction inside `runAdversarialEval` safe.

The registry gains a **third assertion**, because neither of the first two could see this defect: the eval pair was two call sites, neither of them a loop, sending the same rubric — legal under both earlier lists for as long as it existed. It now also asserts which modules send `EVALUATION_RUBRIC`, as an equality, so it fails in both directions.

## 3. #354 and the docs — five corrections, plus one itemised addition

- `instance-setup.md` §5.3 no longer tells an operator to watch `/api/compare` for a trace-gated line, and prints the check (`grep -c trace src/app/api/compare/route.ts` → `0`) beside the claim. It points at `/api/compare-stream`, which takes the same body and does trace.
- §5.3 states the new **45-second per-tool-call ceiling**, names `COMPARE_MCP_TOOL_TIMEOUT_MS`, quotes the message shape, and carries the counterweight.
- `project-plan.md`'s loop sketch says where the loop actually lives.
- `CLAUDE.md`'s model-loop rule states what the registry test does — derived from `git ls-files`, not a two-directory list. Checked against the test, not copied from the phase note.
- `CLAUDE.md`'s two stale rows, **measured at this head**: `# pass 1162` (was 1092) and `3 problems / 3 warnings` (was 4/4).
- **One addition beyond the five**, itemised so it is not mistaken for one of them: a new CLAUDE.md rule recording #323's lesson.

---

## Evidence

Five red demonstrations, plus two more. All output is in the phase report.

| # | Demonstration | Result |
|---|---|---|
| 1 | Uncallable anchored name injected into `CROSS_SOURCE_PREAMBLE` | RED, 2 assertions, naming the constant |
| 2 | …into `SOCRATA_SKILL_FALLBACK` | RED, naming the constant |
| 3 | …into `getSkillForPortal`'s SF output | RED, naming `getSkillForPortal('data.sfgov.org')` |
| 4 | Stale registry entry for the deleted call site | RED — "on the registry but no longer calls" |
| 5 | Still-real registry entry removed | RED — "calls … and is not on the registry" |
| 6 | New guard against base `tools.ts` (criterion 1's literal "adds `fetch`") | RED, naming the fallback **and** `getSkillForPortal` |
| 7 | New guard + base `evaluate/route.ts` (criterion 3's RED) | RED on both the registry and the rubric assertions |

**Suite: `# tests 1162 / # suites 10 / # pass 1162 / # fail 0`** (base: 1158). Name-level diff against the base: **five names added, one removed** — the old preamble-only guard's name. Both of its assertions survive under new names and strictly wider. Full accounting in the phase report.

`npm run build` ✓ · `npm run typecheck` exit 0 ✓ · `npm run lint` `✖ 3 problems (0 errors, 3 warnings)` ✓ · `npm run check:compose-env` PASS ✓ · `npm ci` first.

Every commit carries `Signed-off-by` then `Co-Authored-By`, contiguous, verified after each.

https://claude.ai/code/session_01ADmA5kbRRFfeUJ9hsFvwoJ


npstorey:
**All five checks green on the runner** — run [`33201320982`](https://github.com/npstorey/civic-ai-tools-website/actions/runs/33201320982) at `d83bb25`.

| Check | Result |
|---|---|
| `build / test / lint / typecheck` | pass, 1m16s |
| `container image build` | pass, 1m5s |
| `Vercel` | pass |
| `DCO` | pass |
| `Vercel Preview Comments` | pass |

**The two numbers written into `CLAUDE.md` are confirmed by the runner's own log, not only locally** — which is what criterion 5 is about:

```
# tests 1162
# suites 10
# pass 1162
# fail 0
✖ 3 problems (0 errors, 3 warnings)
```

Base for comparison: `# pass 1158` at `f117665`.

**One accounting item for the gate.** Criterion 6 asks that no test name present at the base be absent. Five names were added and **one was removed**: `every tool name the cross-source preamble advertises exists in mcpTools`, the old preamble-only guard. Both of its assertions survive under new names — its "exactly three `Tools:` lists" structural check is now `#323: the cross-source preamble still describes exactly three sources`, and its callable check is subsumed by `#323: every tool name a prompt-reaching constant presents is callable`, which reads the preamble plus four more constants. No coverage is lost, but a name is gone, and that is reported rather than smoothed over.

**One merge-order item.** `CLAUDE.md`'s `# pass 1162` is measured at this head. P4b is in flight on the same lane; whichever of us merges second makes the other's number stale by the count of the tests it adds. The row should be re-measured at the merged head.

npstorey:
## Two coordinator additions, addressed — `b137ccf`

### 1. `src/lib/mcp/operation-types.ts` — the gap this phase opened, closed

Itemised zone addition, plus its existing test file `src/lib/mcp/operation-types.test.ts` (matches the zone's `src/**/*.test.ts` clause). Diff: `operation-types.ts` +37/−3, its test +88.

**`search` → `'search'`. Added.** `handleSearchTool` takes one property and routes through the same catalog handler `get_data` type=catalog uses (`handleSearchTool` → `handleCatalog`), answering with dataset descriptors. That is the identical shape to the two other name-keyed discovery tools, `search_indicators` and `ckan__search_datasets`, both already `'search'` — one operation, three sources, one label. Second reason it is not `'query'`: some hits carry a few `preview_rows`, and those enrich a catalog hit rather than being rows the analysis queried, so they must stay out of the `records analyzed` sum, which filters on `operationType === 'query'` (#339).

**`fetch` → deliberately absent. This was measured, and the measurement says it cannot be statically typed.**

`handleFetchTool` branches on the **shape of the `id` string at call time**:

- `kind === 'dataset'` → `handleDatasetMetadata`, returns description + columns. **No rows.**
- `kind === 'record'` → `retrieveDocuments`, which fetches from Socrata with a `$where` filter on the detected id field and returns **one real data row**, serialized into `text`.

Nothing in the tool name distinguishes them. The branch is decided by `parseFetchIdentifier`, an identifier grammar with **five accepted forms** — `dataset:…`, `record:…`, a URL classified by its `row_id`/`record_id`/`rowid` query parameters, a bare 4x4 resolved against the server's own default domain, and `4x4:rowid`. Typing `fetch` here would mean reimplementing that grammar in this repository, where it would drift silently — the defect class this wave exists to end — and writing the guess into a **signed** package. Both guesses are wrong in a way a reader cannot see: `'metadata'` hides a genuine record read from `records analyzed`; `'query'` counts every dataset-metadata read as records analyzed. `'unknown'` asserts nothing false.

So `fetch` derives `undefined`, `packager.ts` records `'unknown'`, and **the reason is now in the file** so the next reader does not helpfully fill the gap in. Closing it properly means the server reporting the operation it performed; that is upstream and is not attempted here.

**The property is guarded, not the two instances.** A new test walks every name in `mcpTools` and requires each to resolve to a source *and* to either derive an operation type or sit on a short `UNTYPED_BY_MEASUREMENT` list. It is bidirectional, like the registry.

Two more red demonstrations (nine total for the phase):

| # | Demonstration | Result |
|---|---|---|
| 8 | `fetch: 'metadata'` added — the guess the coordinator asked me to think about | **RED**, 2 assertions: `"fetch" is listed as knowingly untyped but now derives an operation type` |
| 9 | `search: 'search'` removed | **RED**, 2 assertions: `mcpTools advertises "search" and no operation type derives from it. Either map it … or add it to UNTYPED_BY_MEASUREMENT … "unknown" in a signed package should be a decision, not an accident.` |

### 2. `CLAUDE.md`'s test count → **1176**, and it is OBSERVED, not arithmetic

I first wrote this up as arithmetic that "cannot be otherwise, because the merged head does not exist yet." That was wrong, in a phase whose subject is documented claims a reader can falsify, so it is corrected here and in commit `b137ccf` (message-only amend; tree unchanged at `160b326f2fd34b48033e375982b5c50520427336`).

**The merged head does exist.** A `pull_request` run builds the merge ref. Run `33202400885` logged:

```
HEAD is now at ab72cfb Merge 5d0b1f145db725261c614f0cf44b9ab3c3da6b0d into 103de528a7cfc6b736a1a6fda557e69abf37361b
...
# tests 1176
# suites 10
# pass 1176
# fail 0
✖ 3 problems (0 errors, 3 warnings)
```

That is `main` (with P4b) plus this branch — exactly the tree a reader runs the command against. **1176 is measured on a runner at the merged tree.**

The arithmetic agrees and is kept as a cross-check, each term measured:

| Value | Source |
|---|---|
| 1158 at `f117665` (this branch's base) | run `33199189643` |
| 1168 on `main` at `103de52`, after P4b | run `33201542014` — verified myself, not taken on trust |
| 1166 at this head alone | local |

`1158 + 10 + 8 = 1176`. ✔

**One correction to the coordinator's figure**, caused by their own item 1: 1172 was right when my contribution was +4; the `operation-types` tests took it to +8, so it is **1176**.

The lint row stays at **3** — observed on the same merged-tree run.

### Flag added

`streaming.ts` has no `case 'search':` in either `getEducationalAnnotation` or `buildBreadcrumbLabel`, so a `'search'` call gets no educational annotation and falls to `default: return tool.name` for its breadcrumb. **Pre-existing** — `search_indicators` and `ckan__search_datasets` have always behaved this way — but my change makes it a three-tool gap instead of two. Out of zone (`streaming.ts` was P4a's this wave). Fix is one `case` in each of two switches.

npstorey:
Seat GO - head-bound to b137ccf (PR #373), the eighth and last implementing phase. A moved head voids this. Verified by the seat on 2026-08-28 from the remote ref and the runner: three-dot diff +630/-123 over ten files, equal to the PR API; run-tool-loop.ts untouched; all four required checks plus Vercel green on this exact commit; four signed commits, each Signed-off-by then Co-Authored-By, contiguous. Read in the blob: search typed as an operation and fetch deliberately absent with its reason in the file (operation-types.ts:23-39) - the only typing that cannot assert something false today, and the honest-undercount direction for records analyzed; exactly one rubric create (adversarial-eval.ts; the route has none); CLAUDE.md rows measured (# pass 1176 from the merge-ref run, lint 3 problems, the model-loop scope sentence rewritten to git ls-files); instance-setup.md now says the model-identity warning cannot appear on the /api/compare check and why, and documents the 45s ceiling. The merge-ref correction is accepted onto the record: a pull_request run checks out the merge ref, so the merged count is observable before the merge - read it, and let arithmetic be a cross-check. Merge; then P7. Two flags to the close record as follow-on issues: the fallback per-portal search prose (source of truth in the server repo) and the two switches in streaming.ts with no case for search.

PR #379, opened 2026-08-28
fix: what a signed record asserts about identical runs, and what a skeleton URL claims (#363 P8)

npstorey:
Wave N8, phase P8 — chartered after the cold read (G12) on findings F1 and F2. Base `de4bd60`, the merged head of all eight prior phases. Model: **Opus 5** (`claude-opus-5[1m]`).

Two commits, one per task, each carrying `Signed-off-by` then `Co-Authored-By`, contiguous and in order.

---

## Task 1 — the tool-call identity key discriminates on every argument

`45c4ce6`

**The defect, verified at base.** `AttestationDialog.tsx:55-62` keyed a tool call as `name:type:dataset_id:portal`. Ruling D9 made `search` and `fetch` model-callable; their schemas are narrow on purpose (`search` takes only `query`, `fetch` only `id`), so neither carries any of those four fields. Every search collapsed to `search:::`. Two runs that searched for entirely different things produced identical key sets, `jaccardSimilarity` returned 1, `(toolCallOverlap + outputSimilarity)/2 >= 0.9` classified them `highly_reproducible`, and `:303` wrote those keys into a **signed** consistency attestation whose overlap renders as "Tool overlap: 100%" (`AttestationSection.tsx:362`).

**The fix is the property, not a longer list.** The key is the tool name plus an RFC 8785 canonical serialisation of the whole argument object. Canonical rather than raw, because arguments arrive as a parse of the endpoint's own key order: two identical calls written in different orders must stay one key, since scoring a reproducible run as *inconsistent* is just as false a claim to sign.

**One implementation.** The canonicalisation moved to `src/lib/evidence/tool-call-identity.ts`; `AttestationDialog.tsx` and `replay-loop.test.ts` both import it, and the verbatim copy at `replay-loop.test.ts:152-159` is gone. That copy existed because no `.test.ts` in this tree imports a `.tsx`, and its own comment argued a shared helper "would quietly agree with itself" — it was the copy that agreed with itself, on both of the failures above.

**Both REDs, shown red against the shipped behaviour before the fix.** The shipped function was extracted verbatim into the new module first, so the REDs ran against the real code rather than a description of it.

RED 1 — two different searches:
```
not ok 1 - two searches for different phrases are two different keys
  error: 'a search key must carry the phrase that was searched for'
  expected: 'search:::'
  actual:   'search:::'
  operator: 'notStrictEqual'

not ok 3 - two runs that searched for different things do not have identical key sets
  error: two runs that read different data must not present as one repeated run
         true !== false
```

RED 2 — the same dataset, a different `where` (the half the wave did not cause; `where` was never in the key):
```
not ok 4 - the same dataset with different WHERE clauses is two different keys
  error: 'a filter that changes the rows must change the key'
  expected: 'get_data:query:erm2-nwe9:data.cityofnewyork.us'
  actual:   'get_data:query:erm2-nwe9:data.cityofnewyork.us'
  operator: 'notStrictEqual'
```
9 of 13 red before, 13 of 13 green after.

**Historical attestations are untouched.** `toolCallKeys` appears exactly once in the tree (`AttestationDialog.tsx:298`, the write site) — nothing reads it back by name. Both production call sites take `completedRuns`, accumulated from live `POST /api/records/:slug/replay` responses; the route hashes and signs the submitted body verbatim; `AttestationSection` renders the stored `metrics.toolCallOverlap` and fetches the stored blob for the expanded view; `attestation-backfill.ts:250` signs `row.packageHash` as stored. No path recomputes a stored attestation.

**Criterion 14's pin is superseded, as the contract directs.** `replay-loop.test.ts`'s `the attestation identity keys are unchanged, injected portal included` is renamed to `the attestation identity keys carry the injected portal` and its expected keys updated to the new format. What the pin protects — that the portal the loop core injects into the recorded args survives into the keys — is unchanged and still asserted end-to-end through the real loop. **This is the one base test name absent from the new run, and it is the one the contract permits.**

---

## Task 2 — a skeleton notebook never emits an undefined URL, and carries every argument

`cd71d0d`

**The defect, verified at base by rendering.** `notebook.ts:59-72` built `https://${portal}/resource/${datasetId}.json` with no guard and handled only select/where/group/order/limit. Its step filter (`:130`) is `operationType === 'query'`, which since `operation-types.ts` also admits `get_observations`, `ckan__query_data`, `ckan__execute_sql` and `ckan__aggregate_data` — none of which carry a portal or dataset id. `PublishEvidenceDialog.tsx:167` puts this notebook into the signed package.

Criterion 4's RED, produced by rendering `generateNotebook` at `de4bd60`:
```
--- markdown ---
## Step 1: Median income

*Original query returned 47 rows, 3 columns.*
--- code ---
url = "https://undefined/resource/undefined.json"

response = requests.get(url)
...
--- markdown ---
## Step 2: Complaint counts

*Original query returned 5 rows, 2 columns.*
--- code ---
url = "https://data.cityofnewyork.us/resource/erm2-nwe9.json?$where=borough%3D'QUEENS'&$limit=10"
```
Step 2 is criterion 5's RED in the same render: the call carried `query: "SELECT complaint_type, count(*) GROUP BY complaint_type LIMIT 5"` and `offset: 100`, both dropped in silence, while `$where` and `$limit` — which the data source sets aside under a `$query` — were written as parameters with no effect. The cell returns a different page than the analysis read, under the analysis's row count. Worse than the #340 it mirrors: #340 raised a `TypeError` and told the reader live data could not be fetched; this runs clean and is confidently wrong.

The same fixture after the fix:
```
--- markdown ---
## Step 1: Median income

*Original query returned 47 rows, 3 columns.*

*Not reproduced below.* This step called `get_observations`, and the cells in this
notebook fetch Socrata datasets by URL. This call named no Socrata portal and
dataset, so no URL is written for it — one written here would name a source the
step did not read. The original result is described above.
--- markdown ---
## Step 2: Complaint counts

*Original query returned 5 rows, 2 columns.*
--- code ---
# `query` here is a full SoQL statement, so the data source applies it as the
# whole query ($query): select / where / group / order — and limit / offset —
# are not sent alongside it. They are omitted from the URL rather than written
# as parameters that would have no effect on the rows this cell returns.
# Superseded on this call, for that reason: where, limit, offset.
# The row count is therefore bounded by the statement's own LIMIT, or by the
# portal's default page size when the statement carries none.
url = "https://data.cityofnewyork.us/resource/erm2-nwe9.json?$query=SELECT%20complaint_type%2C%20count(*)%20GROUP%20BY%20complaint_type%20LIMIT%205"
```

**What a step with no URL renders, and why.** It keeps its number and its result summary and gains a note naming the operation it could not reproduce. Silence was rejected: dropping the step would make the document claim the analysis rested on fewer sources than it did — the same false precision (design-principles.md Principle 3) pointing the other way, and against Principle 2's persistent source marking. The note names the tool and stops there, so it claims no source it did not read.

**Every advertised argument accounted for.** `portal`/`dataset_id` are the path; `type` is routing metadata that already selected this renderer; `query` follows the data source's precedence — SoQL-shaped becomes `$query` alone with no `$limit`/`$offset` beside it, a phrase becomes `$q` alongside the clauses; `select`/`where`/`group`/`order`/`limit`/`offset` become their `$` parameters. Superseded clauses and any argument this URL has no parameter for are named in the cell rather than dropped without a trace. `limit: 0` is now a limit rather than a falsy absence, and a clause whose value the URL cannot express (anything but a string or a finite number) is disclosed rather than stringified into it.

**The SoQL sniff is imported, not copied.** `notebook.ts` imports `isFullSoqlQuery` from `notebook-author/tool-to-cell.ts` — the function whose body is `SOQL_QUERY_SNIFF.test(query)`. No fourth copy, and `tool-to-cell.ts` is unmodified. Rationale and the alternative are in the report.

**The "Data sources" list** is now built from the planned steps, so it cites only datasets a cell in the document actually fetches, on the portal that cell fetches them from. It previously read `dataset_id` off any query call and fell back to the run's selected portal for the link.

---

## Gates

`npm ci` first. All five green; full output in the phase report.

| Gate | Result |
|---|---|
| `npm test` | `# tests 1205 / # suites 10 / # pass 1205 / # fail 0` (base `de4bd60`: 1176) |
| `npm run build` | `✓ Compiled successfully`, Route (app) table |
| `npm run lint` | `✖ 3 problems (0 errors, 3 warnings)` — the documented baseline |
| `npm run check:compose-env` | `RESULT: PASS — every variable this profile reads can reach the container.` |

`+29` tests, `-0`, one rename. Green on the runner: run `33207707041`, checks `build / test / lint / typecheck` and `container image build`, both **pass**; `DCO` and `Vercel` pass. The two Turbopack `Dynamic filesystem access` warnings are pre-existing — measured at `de4bd60` on this machine, same count, and their import traces run `helpers/index.ts → synthesize.ts → {query-notebook route, notebook-preview page}`, which this diff does not touch.

## Blast zone

Six files, all inside the declared zone:

```
 src/components/evidence/AttestationDialog.tsx |  15 +-
 src/lib/evidence/tool-call-identity.test.ts   | 184 +++++++++++++++++++
 src/lib/evidence/tool-call-identity.ts        |  64 +++++++
 src/lib/model-loop/replay-loop.test.ts        |  71 ++++----
 src/lib/notebook.test.ts                      | 245 +++++++++++++++++++++++++
 src/lib/notebook.ts                           | 253 +++++++++++++++++++++++---
 6 files changed, 767 insertions(+), 65 deletions(-)
```

`tool-to-cell.ts` was **not** modified — the shared sniff needed neither to move nor to be re-exported. Nothing outside the zone was touched; three out-of-zone findings are flagged in the phase report, not fixed.

### DIFF
--- a/src/components/evidence/AttestationDialog.tsx
+++ b/src/components/evidence/AttestationDialog.tsx
@@ -2,6 +2,11 @@
 
 import { useState, useCallback, useEffect } from 'react';
 import { parseModelsResponse, type Model } from '@/lib/model-list';
+// The tool-call identity key, in ONE place. It used to be a private function
+// here with a verbatim copy in `replay-loop.test.ts`, because no `.test.ts` in
+// this tree can import a `.tsx` — so the shipped function had no test and the
+// copy could only ever agree with itself. Both now import this module.
+import { canonicalizeToolCall } from '@/lib/evidence/tool-call-identity';
 
 // --- Types ---
 
@@ -51,16 +56,6 @@ interface AttestationDialogProps {
 }
 
 // --- Metric computation ---
-
-function canonicalizeToolCall(tc: { name: string; args: Record<string, unknown> }): string {
-  const key = [
-    tc.name,
-    tc.args.type as string || '',
-    tc.args.dataset_id as string || '',
-    tc.args.portal as string || '',
-  ].join(':');
-  return key;
-}
 
 function jaccardSimilarity(a: Set<string>, b: Set<string>): number {
   const intersection = new Set([...a].filter(x => b.has(x)));
--- a/src/lib/evidence/tool-call-identity.test.ts
+++ b/src/lib/evidence/tool-call-identity.test.ts
@@ -1,0 +1,184 @@
+// The tool-call identity key a signed consistency attestation is built from.
+//
+// WHY THIS FILE EXISTS. The key used to be a private function inside
+// `AttestationDialog.tsx`, and a verbatim copy of it lived in
+// `replay-loop.test.ts` under a comment saying the copy going stale would be
+// "the loud failure" — because no `.test.ts` in this tree imports a `.tsx`, so
+// the real function had no test at all. Two implementations that can drift is
+// not a guard; it is the defect wearing a guard's clothes. The function now
+// lives in `./tool-call-identity.ts`, the dialog and the tests both import it,
+// and this file is the only place its behaviour is asserted.
+//
+// THE PROPERTY. Any difference in a tool call's arguments must produce a
+// different key. The old key was a hand-picked field list
+// (`name:type:dataset_id:portal`), and a hand-picked list is exactly what
+// failed — twice over: `search` and `fetch` carry none of those four fields,
+// and `where` was never in the list even for `get_data`. Both cases are below.
+//
+// Run with: npm test
+//   (or: node --test --experimental-strip-types src/lib/evidence/tool-call-identity.test.ts)
+
+import { test } from 'node:test';
+import assert from 'node:assert/strict';
+import { canonicalizeToolCall } from './tool-call-identity.ts';
+
+/** A replay run, as `AttestationDialog` reads it off `/api/records/:slug/replay`. */
+type Run = { name: string; args: Record<string, unknown> }[];
+
+/** The key SET of a run — what `computeConsistencyMetrics` compares pairwise. */
+const keysOf = (run: Run): Set<string> => new Set(run.map(canonicalizeToolCall));
+
+const sameSet = (a: Set<string>, b: Set<string>): boolean =>
+  a.size === b.size && [...a].every(k => b.has(k));
+
+// --- RED 1: two different searches -----------------------------------------
+//
+// Ruling D9 made `search` and `fetch` model-callable. Their schemas are narrow
+// on purpose: `search` takes only `query`, `fetch` only `id`. Neither carries
+// `type`, `dataset_id` or `portal`, so under the old key every search
+// collapsed to `search:::` and every fetch to `fetch:::`. Two runs that
+// searched for entirely different things produced identical key sets, Jaccard
+// returned 1, and `(1 + outputSimilarity)/2 >= 0.9` classified them
+// `highly_reproducible` — in a SIGNED consistency attestation rendering
+// "Tool overlap: 100%".
+
+test('two searches for different phrases are two different keys', () => {
+  const noise = canonicalizeToolCall({ name: 'search', args: { query: 'noise complaints' } });
+  const rats = canonicalizeToolCall({ name: 'search', args: { query: 'rat sightings' } });
+  assert.notEqual(noise, rats, 'a search key must carry the phrase that was searched for');
+});
+
+test('two fetches of different identifiers are two different keys', () => {
+  const a = canonicalizeToolCall({ name: 'fetch', args: { id: 'dataset:data.cityofnewyork.us:erm2-nwe9' } });
+  const b = canonicalizeToolCall({ name: 'fetch', args: { id: 'dataset:data.cityofnewyork.us:zzzz-9999' } });
+  assert.notEqual(a, b, 'a fetch key must carry the identifier that was fetched');
+});
+
+test('two runs that searched for different things do not have identical key sets', () => {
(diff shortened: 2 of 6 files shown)


npstorey:
Seat GO - head-bound to cd71d0d (PR #379), the fix-on-top phase the cold read earned; the last phase of Wave N8. A moved head voids this. Verified by the seat on 2026-08-28 from the remote ref and the runner: three-dot diff +767/-65 over six files, equal to the PR API; all four required checks plus Vercel green on this exact commit; two signed commits, each Signed-off-by then Co-Authored-By, contiguous. Read in the blob: the identity key is now name plus jcs(args) (src/lib/evidence/tool-call-identity.ts:62-63, the repository own RFC 8785 canonicalizer), the dialog imports it, and toolCallKeys has one write site (AttestationDialog.tsx:298) that reads live replay responses; the tests pin both REDs the ruling named (two different searches; same dataset, different WHERE) plus argument-order stability and the tool-name discriminator. notebook.ts now refuses to address a URL it cannot honestly build and carries every Socrata clause the call sent (the $query path documented at :179 and :215). Merged count read from the merge-ref run: 1205 (1176 + 29) is the tree this merge produces; arithmetic agrees. Merge; cut the -merged tag; file the stale-format references as their own issue; then the close record.

